PR #16, opened 2022-07-05
various makefile and ocamlformat tweaks + fix CI

Alizter:


Makefile
```diff
@@ -2,13 +2,21 @@
...
+DEV_DEPS= \
+ocamlformat.0.22.4 \
+ocaml-lsp-server
 
 build: coq_boot
 	dune build $(DUNEOPT) $(PKG_SET)
 
-format: coq_boot
-	dune fmt
+fmt format: coq_boot
```
ejgallego, line 18: This is a PHONY target , right? Let's declare it then.

PR #32, opened 2022-07-11
[lsp] Request Coq interrupt when we receive a new check request.

ejgallego:
More testing for this is needed, but that's a nice first step, as
actually setting the flag will eventually raise an async exception
which is what we want in the general case.

Note however that we must indeed handle the async interruption in our
code properly, in particular we MUST NOT cache the user-interrupt,
that is to say, we need to make the memo cache clean up properly on
the async interrupt.

### DIFF
--- a/controller/coq_doc.ml
+++ b/controller/coq_doc.ml
@@ -129,7 +129,8 @@ let process_and_parse ~ofmt ~uri ~version ~coq_queue doc =
       match parse_stm ~st doc_handle with
       | Ok None, time -> (EOF, diags, time)
       | Ok (Some ast), time -> (Process ast, diags, time)
-      | Error (loc, msg), time ->
+      | Error Coq_util.Error.Interrupted, time -> (Skip, diags, time)
+      | Error (Coq_util.Error.Eval (loc, msg)), time ->
         let diags = (to_orange loc, 1, to_msg msg, None) :: diags in
         discard_to_dot doc_handle;
         (Skip, diags, time)
@@ -188,7 +189,10 @@ let process_and_parse ~ofmt ~uri ~version ~coq_queue doc =
         let node = { ast; exec = true; goal = pr_goal st } in
         let doc = { doc with nodes = node :: doc.nodes } in
         stm doc st diags
-      | Error (err_loc, msg) ->
+      | Error Coq_util.Error.Interrupted ->
+        (* Exit *)
+        (doc, st, diags)
+      | Error (Coq_util.Error.Eval (err_loc, msg)) ->
         let loc = Option.append err_loc loc in
         let diags = (to_orange loc, 1, to_msg msg, None) :: diags in
         let node = { ast; exec = false; goal = pr_goal st } in
--- a/controller/coq_interp.ml
+++ b/controller/coq_interp.ml
@@ -5,7 +5,7 @@ module Info = struct
     }
 end
 
-type 'a interp_result = ('a Info.t, Loc.t option * Pp.t) result
+type 'a interp_result = ('a Info.t, Coq_util.Error.t) result
 
 let coq_interp ~st cmd =
   let st = Coq_state.to_coq st in
@@ -22,7 +22,8 @@ let marshal_out f oc res =
   | Ok { Info.res; warnings = _ } ->
     Marshal.to_channel oc 0 [];
     f oc res
-  | Error (loc, msg) ->
+  | Error Coq_util.Error.Interrupted -> ()
+  | Error (Eval (loc, msg)) ->
     Marshal.to_channel oc 1 [];
     Marshal.to_channel oc loc [];
     Marshal.to_channel oc msg [];
@@ -36,4 +37,4 @@ let marshal_in f ic =
   else
     let loc : Loc.t option = Marshal.from_channel ic in
     let msg : Pp.t = Marshal.from_channel ic in
-    Error (loc, msg)
+    Error (Coq_util.Error.Eval (loc, msg))
--- a/controller/coq_interp.mli
+++ b/controller/coq_interp.mli
@@ -5,7 +5,7 @@ module Info : sig
     }
 end
 
-type 'a interp_result = ('a Info.t, Loc.t option * Pp.t) result
+type 'a interp_result = ('a Info.t, Coq_util.Error.t) result
 
 val interp : st:Coq_state.t -> Coq_ast.t -> Coq_state.t interp_result
 val marshal_in : (in_channel -> 'a) -> in_channel -> 'a interp_result
--- a/controller/coq_util.ml
+++ b/controller/coq_util.ml
@@ -1,9 +1,17 @@
+module Error = struct
+  type t =
+    | Eval of Loc.t option * Pp.t
+    | Interrupted (* signal sent, eval didn't complete *)
+end
+
 let coq_protect ~f x =
   try
     let res = f x in
     Ok res
-  with exn ->
+  with
+  | Sys.Break -> Error Error.Interrupted
+  | exn ->
     let e, info = Exninfo.capture exn in
     let loc = Loc.(get_loc info) in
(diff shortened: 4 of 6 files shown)


ejgallego:
I suggest you update the type of coq_protect as to return `Interrupted` status, then the type system will indicate all the places to patch.

ejgallego:
I did it, but needs some review / testing.

Alizter:
Seems to work pretty well! When checking Pff.v, it is interesting to see that the document is being checked from the beginning every time you edit a line. I don't think this is new however.

ejgallego:
We don't have yet caching of parsing, so that's expected. Or do you see the interpretation cache missing too in Pfff 

I have a patch for caching parsing tho, that should help a lot in that case.

Note that if you enabled continuous diagnostics then you have a O(n^2) message passing cost. We need to open an issue with LSP upstream to request streaming of diagnostics.

ejgallego:
We should profile the cost of sending the diagnostics by the way, it is far from trivial in cases like Pfff. There, we should actually implement streaming internally too, I bet that by itself speeds up things by quite a bit in Pfff.

Alizter:
> Note that if you enabled continuous diagnostics then you have a O(n^2) message passing cost. We need to open an issue with LSP upstream to request streaming of diagnostics.

I still think we are using the wrong solution. We want both highlights and underlining right? Then we should be using the dedicated highlighting mechanism in LSP (I forget what it is called exactly). For the underlining, I still think there might be other ways of achieving what we are after.

Shoving it all through diagnostics seems like abuse to me.

Alizter (approved):

Alizter:
The cache is mostly a miss it seems.

ejgallego:
> The cache is mostly a miss it seems.

Strange, what's the hit number you see in the log? What line are you editing?

Alizter:
I think it was around 1000 of Pff.v

ejgallego:
> I still think we are using the wrong solution. We want both highlights and underlining right? Then we should be using the dedicated highlighting mechanism in LSP (I forget what it is called exactly). For the underlining, I still think there might be other ways of achieving what we are after.
> 
> Shoving it all through diagnostics seems like abuse to me.

Sure, it would be great to implement a better mechanism. It is a bit of an abuse, but has gotten us quite far (on all platforms), so for a prototype it is still a good tradeoff. What are these other ways?

Anyways, the main point in my comment is not about the concrete mechanism, but about the problem we face due to lack of streaming notifications in LSP, this is a problem for any mechanism we choose, and we should discuss that with upstream.

Alizter:
>Sure, it would be great to implement a better mechanism. It is a bit of an abuse, but has gotten us quite far (on all platforms), so for a prototype it is still a good tradeoff. What are these other ways?

It is called hover, we can for each client implement highlighting if we want that, but these are separate concerns. The issue is that we don't want to spam the protocol with the diagnostics, and hover will allow for the user to request what they want to see.

I tried experimenting with it yesterday but we will need some more LSP stuff. The ocaml-lsp library has all the bindings ready for the json stuff, and I was hoping to vendor and use that. Ultimately it didn't come to fruition since Coq has a Pp library which name clashes with the one from ocaml-lsp and it gave me an OCaml headache.

I still think it would be good to use the ocaml-lsp "lsp" library. I'm not talking about other LSP features, but simply the types they have already defined for describing whats in the LSP spec.

ejgallego:
Note that we already implement hover, but how can you do highlighting with that?

I think I'd be good to use the OCaml lsp library, but I'm confused as why we need it now, in particular for hover, as we already implement it. What concrete stuff are you missing from there?

ejgallego:
> There is a separate hover method in LSP which lets you hover on demand. This means we wouldn't be sending all the diagonostics but only the ones requested and the errors.

Yes, this method is already implemented.



> What I meant with highlighting is that it should be done on a per client basis. The fact that the info diagnostic highlights a line should not be why we use it. When we have more than 100 diagnostics being sent, we start to pay the price in performance.

What is a "per-client" basis. How does that avoid sending a check status for all the things that checked OK?

> If we start using it now, it will be easier to transition then further in the future. To be clear, I'm not talking about ocaml-lsp as a whole, but simply the bindings for the json stuff. It also saves us from having to reimplement stuff.

Yup, but I'm curious what you need from the lib. It'd be suprised you need anything for hover as I already implemented it some time ago.

Alizter:
> > There is a separate hover method in LSP which lets you hover on demand. This means we wouldn't be sending all the diagonostics but only the ones requested and the errors.
> 
> Yes, this method is already implemented.
> 

I'm talking about this: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_hover

Is this really implemented?

> > What I meant with highlighting is that it should be done on a per client basis. The fact that the info diagnostic highlights a line should not be why we use it. When we have more than 100 diagnostics being sent, we start to pay the price in performance.
> 
> What is a "per-client" basis. How does that avoid sending a check status for all the things that checked OK?
> 

Sorry I meant a per-client-request basis. The hover request I linked above is done by the client and the server responds. As opposed to the diagnostics which are constantly published from the server to the client. 

> > If we start using it now, it will be easier to transition then further in the future. To be clear, I'm not talking about ocaml-lsp as a whole, but simply the bindings for the json stuff. It also saves us from having to reimplement stuff.
> 
> Yup, but I'm curious what you need from the lib. It'd be suprised you need anything for hover as I already implemented it some time ago.

I'm specifically talking about this area of the lib: https://github.com/ocaml/ocaml-lsp/tree/master/lsp/src

Take for instance https://github.com/ocaml/ocaml-lsp/blob/master/lsp/src/types.mli

We have some of these in our lsp folder, but we don't need to redo the work already done here.

ejgallego:
> Is this really implemented?

Yes, it is pretty obvious from the code, isn't it? Actually implementing hover in a proper way requires some interesting data structures, lambdapi has a prototype that we should copy at some point.

But in a sense this is why I think centering the discussion about external libraries in a project like this does distract us from the core topics, like what is missing from our hover implementation, and what we need to evolve it (and I'm sure that unfortunately ocaml-lsp cannot help in any way here)

Anyways I guess you can experiment with hover as is now and provide more feedback about hover in a separate issue.

Alizter:
> > Is this really implemented?
> 
> Yes, it is pretty obvious from the code, isn't it? Actually implementing hover in a proper way requires some interesting data structures, lambdapi has a prototype that we should copy at some point.

No, it is not obvious to me. I don't see anything about hover in lsp/base.ml. I don't think that the data structures are particularly interesting either, it's just that I am not interested in writing them again especially if somebody else has done them generically for the entire protocol. I'm also not sure what prototype you are talking about.

> But in a sense this is why I think centering the discussion about external libraries in a project like this does distract us from the core topics, like what is missing from our hover implementation, and what we need to evolve it (and I'm sure that unfortunately ocaml-lsp cannot help in any way here)

You mentioned earlier that sending diagonostics are becoming a bottleneck. I am trying to propose an alternative solution. Part of this solution requires using parts of the LSP protocol we don't use currently. I merely suggested that we can use existing work on this rather than reimplementing the LSP datastructures ourselves.

ejgallego:
We follow a different implementation structure, methods are implemented in coq_lsp.ml , in particular `do_hover`.

I agree that it would be good not to reimplement data structures, however that is far from being the bottleneck here.

The lambdapi prototype is the one referenced in the readme https://github.com/Deducteam/lambdapi/tree/master/src/lsp , the parent of coq-lsp.



> You mentioned earlier that sending diagonostics are becoming a bottleneck. I am trying to propose an alternative solution.

Diagnostics as implemented now are O(n^2) , what I mean is that any other solution with the same functionality will be also O(n^2) unless we extend the protocol to stream data.

The rest doesn't follow as you are assuming that coq-lsp doesn't implement hover, but it does (and it can't solve the same case than diagnostics by definition)

PR #34, opened 2022-07-12
[fleche] When interrupted in parsing, signal end of file

ejgallego:
I think that's the correct thing to do, as we don't want to continue
as of now with error recovery, etc...


Alizter:
@ejgallego can you format

PR #36, opened 2022-07-13
Memo parsing

ejgallego:


controller/memo.ml
```diff
@@ -102,6 +102,82 @@ let interp_command ~st stm : _ Stats.t =
 
 let mem_stats () = Obj.reachable_words (Obj.magic cache)
 
+module ParseInput = struct
+  type t = Pvernac.proof_mode option * Vernacstate.Parser.t * Loc.t * string
+
+  let hash x = Hashtbl.hash x
```
Alizter, line 108: We shouldn't be hashing locations right?

ejgallego: As far as I can see, without using really advanced incremental parsing trickery, when the location of a sentence moves, we need to reparse, otherwise the locations on the AST would be wrong.

In fact, now that I think of it, ignoring locations on the AST is not fully sound for interpretation, as cached messages would get the wrong location then. So to that extend, we need to eventually introduce a more refined command type, and then if a command produced or stored output with locations, the command could need re-run.

That's quite tricky acutally :( :(

ejgallego:
Now that we can resume correctly this doesn't seem to be too valuable, closing for now.

PR #37, opened 2022-07-27
[serapi] Bump to version that implements hash / compare for tactics.

ejgallego:
This makes the cache much more efficient, with now almost perfect hit
rate.

However, this also makes obvious the problem we have with our caching
system as indeed located messages (and diagnostics) are unsound
w.r.t. to location-independent caching.

We must implement a proper fix to this strategy before merging this; I
think we can indeed make the cache aware of the location a command was
cached, and thus include the offset when we get a hit, and adjust
the cached messages accordingly. Exciting!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,6 @@
 - Location-aware cache for Coq interpretation (@ejgallego)
 
 # coq-lsp 0.0.0.1
--------
+-----------------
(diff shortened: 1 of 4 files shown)


ejgallego:
Memo is now aware of location offset and errors are corrected in the same way.

PR #41, opened 2022-10-01
[info] Hook completion call, fix still some issues with range.

ejgallego:


Alizter:
What should we do with this? Shall I review?

ejgallego:
Needs progress @Alizter , the way we extract the information from the buffer needs to be more structured, as proposed in #368 for example.

Feel free to make progress on #368 tho

PR #42, opened 2022-10-02
[coq] Add plugin loader library.

ejgallego:
This is just a copy from SerAPI's one for now, but it belongs here.

### DIFF
--- a/coq/loader.ml
+++ b/coq/loader.ml
@@ -1,0 +1,51 @@
+(************************************************************************)
+(*         *   The Coq Proof Assistant / The Coq Development Team       *)
+(*  v      *   INRIA, CNRS and contributors - Copyright 1999-2018       *)
+(* <O___,, *       (see CREDITS file for the list of authors)           *)
+(*   \VV/  **************************************************************)
+(*    //   *    This file is distributed under the terms of the         *)
+(*         *     GNU Lesser General Public License Version 2.1          *)
+(*         *     (see LICENSE file for the text of the license)         *)
+(************************************************************************)
+
+(************************************************************************)
+(* Coq serialization API/Plugin                                         *)
+(* Copyright 2016-2019 MINES ParisTech                                  *)
+(* Copyright 2019-2022 Inria                                            *)
+(* Written by Emilio J. Gallego Arias                                   *)
+(************************************************************************)
+[@@@ocamlformat "disable"]
+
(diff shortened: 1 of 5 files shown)


coq/loader.ml
```diff
@@ -0,0 +1,65 @@
...
+    | "coq-core.plugins.ssreflect"        (* ssr *)
+      -> true
+    | _ ->
+      Feedback.msg_warning Pp.(str "Missing serlib plugin: " ++ str fl_pkg);
+      false
+  in
+  if supported
+  then
+    let plugin_name = String.split_on_char '.' fl_pkg |> list_last in
+    Some ("coq-serapi.serlib." ^ plugin_name)
```
Alizter, line 41: Does this need to be renamed?

coq/loader.ml
```diff
@@ -0,0 +1,65 @@
...
+          Format.eprintf "[plugin loader]: plugin %s requested via findlib@\n%!" fl_pkg;
+        loader [serlib_pkg]
+      | None ->
+        if debug then
+          Format.eprintf "[plugin loader]: plugin %s requested via mltop@\n%!" fl_pkg;
+        loader [fl_pkg]
+    with
+    | Dynlink.Error err as exn ->
+      let msg = Dynlink.error_message err in
+      Format.eprintf "[sertop] Critical Dynlink error %s@\n%!" msg;
```
Alizter, line 61: Also this

coq/loader.ml
```diff
@@ -0,0 +1,65 @@
...
+      raise exn
+    | Fl_package_base.No_such_package (pkg, _) as exn ->
+      Format.eprintf "[sertop] Couldn't find the SerAPI plugin %s@; please check `ocamlfind list` does include SerAPI's plugin libraries.\n%!" pkg;
```
Alizter, line 64: and this

coq/loader.ml
ejgallego: No, it does not need renaming. That's a tricky issue as plugins can't register themselves before they are loaded, so for now a manual list needs to be kept.

ejgallego: Likely we will add a static init to serlib so at least the list is local.

ejgallego: "No it does not need renaming" really means "no, renaming is wrong and would break everything"

PR #45, opened 2022-10-06
tweak wording in editor/code

Alizter:

ps-id: a19c551e-7427-4557-a4d2-816345146880


Alizter:
OK let me have another look.

ejgallego:
The code looks bizarre now, `start` was named fine, the problem was in the messages, they are for two related components.

editor/code/src/client.js
```diff
@@ -3,36 +3,40 @@
 const vscode = require("vscode");
 const vscc = require("vscode-languageclient");
 
-function activate (context) {
-    vscode.window.showInformationMessage('Going to activate!');
+function activate(context) {
+    vscode.window.showInformationMessage('Starting coq-lsp!');
```
ejgallego, line 7: This should be something like "loading coq-lsp mode"

editor/code/src/client.js
```diff
@@ -3,36 +3,40 @@
...
         documentSelector: [
-            {scheme: 'file', language: 'coq'}
+            { scheme: 'file', language: 'coq' }
         ]
     };
 
     let client = null;
 
-    const restart = () => {
+    const start = () => {
```
ejgallego, line 16: This really does restart, so the old name was better.

editor/code/src/client.js
```diff
@@ -3,36 +3,40 @@
...
-    vscode.commands.registerCommand('vsc-galinas.restart', restart);
 
-    restart();
+    const restart = () => {
+        vscode.window.showInformationMessage('Restarting coq-lsp!');
+        start();
+    };
+
+    vscode.commands.registerCommand('vsc-galinas.restart', restart);
+    start();
```
ejgallego, line 39: Now the useful information about the server starting is gone.

editor/code/src/client.js
Alizter: I thought it was silly for it to restart as the first thing it does.

Alizter:
Oh well, lets tweak the wording some other time.

PR #47, opened 2022-10-06
[build] Also build ltac2 in the composed setup.

ejgallego:
This seems to be problematic for the CI, seems like a mystery to me.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,8 @@ coq/config/coq_config.ml:
 	cd vendor/coq \
 	&& ./configure -no-ask -prefix $(shell pwd)/_build/install/default/ \
 		-native-compiler no \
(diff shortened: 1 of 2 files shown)


ejgallego:
Of course the runners were picking the buggy 3.4 Dune version, which cant' build our setup.

Alizter:
@ejgallego I feel like this will keep biting. Can we ask for a 3.4.1 backport of your patch?

ejgallego:
> @ejgallego I feel like this will keep biting. Can we ask for a 3.4.1 backport of your patch?

Sure, tho with 3.5 so close I'm not sure if that'll help a lot. We'd have to add a conflict with 3.4.0 anyways so that'd be a PITA.

PR #58, opened 2022-10-11
[debug] make OK dialogue a configurable debug option

Alizter:
I've set the default to off for now. I think it should be upto the clients to show a green background on processed things since LSP doesn't have such an option.

ps-id: ef7beba0-9698-44ae-a9dc-5b0922cf64cf


fleche/debug.ml
```diff
@@ -13,3 +13,6 @@ let parsing = false || all
 
 (* Backtraces *)
 let backtraces = false || all
+
+(* Show diagnostic for OK lines *)
+let ok_diag = false || all
```
ejgallego, line 18: This needs to go in `config.ml`

Alizter: Done

Alizter:
@ejgallego That should be straight forward.

PR #60, opened 2022-10-11
mark vendor/ as vendored_subdir for Dune

Alizter:
This stops things like ocamlformat or compiler warnings from being checked in the vendored directories.

ps-id: 7a8b280a-bc4d-45e2-a966-b28eddb2fde8


ejgallego (approved):
Thanks!

PR #62, opened 2022-10-13
[fleche] Basic markdown support

ejgallego:
Ported from jsCoq; .mv files will be handled as literate markdown files.


Alizter:
Very cool!

ejgallego:
Hehe, dirty but effective trick :) Please feel free to:
- add some example from Alectryon
- improve the syntax highglighting (I'm unsure how to do this tho in vscode)

At some point we could extend the editor to better recognize the markdown, but this is hard.

PR #81, opened 2022-11-28
chore: Nix detects .ocamlformat version + CI fmt uses Nix

Alizter:
This gives a faster lint job and allow the flake to follow the version of ocamlformat seamlessly.


Alizter:
failure unrelated

Issue #51, opened 2022-10-07
Build setup


Alizter:
1. Remove vendored submodules
2. Use coq-universe as a playground for examples

ejgallego:
Actually coq-universe would be the main developer workspace then.

ejgallego:
So far this is a big problem and the only solution I could think of is as follows:

For now, the tree setup (in jscoq, here) looks like:
```
Makefile
dune-project
....
fleche/
vendor/coq
vendor/coq-serapi
```

but ideally the setup would look like
```
workspace-config.yml
Makefile
coq-lsp
coq
coq-serapi
```

where coq-lsp knows nothing about what's vendored, but some tool will read workspace-config.yml , checkout the projects at the right version, and setup a Makefile.

Alizter:
Actually, having a vendor/ directory where a user can plop symlinks to their own projects in would be quite a good workflow.

ejgallego:
Yes, in jsCoq we do something similar, so what we are leaning towards is a system where the submodules are optional.

PR #82, opened 2022-11-28
chore(ci): add Nix to CI

Alizter:
I've also removed some duplication between the macos and ubuntu jobs since we can be specific about combinations we actually want to check.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,6 +48,27 @@ jobs:
       - name: 🧱 Build coq-lsp
         run: opam exec -- make build
 
+  build-nix:
+    name: Nix
+    strategy:
+      matrix:
+        os: [macOS-latest, ubuntu-latest]
+      fail-fast: false
+
+    runs-on: ${{ matrix.os }}
+
+    steps:
+      - name: 🔭 Checkout code
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - name: ❄️ Setup Nix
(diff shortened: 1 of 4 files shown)


ejgallego:
Hi @Alizter , what's the milestone for this and #85 ?

Alizter:
@ejgallego issues have been fixed. I've reduced the dune bound in the opam file to 3.2. This can be merged ASAP.

ejgallego (changes requested):
This needs changes:
- dune < 3.5 will make CI fail, so the bound is still needed as long as we have the current setup, bump Nix stuff instead to use 3.5
- please put unrelated changes in their own commits / PRs

Alizter:
@ejgallego the CI seems to be fine with the version, are you sure?

ejgallego:
CI seems fine because it is picking a more up to date version, but if we relax the constraint it would accept non-working configurations.

Indeed we have a problem with the way things are setup, (issue #51 ) , for a non-composed build, we can indeed use Dune 3.2 or lower (see opam file for the v8.16 branch) , however, for the composed build, we need Dune >= 3.5 .

I'm still unsure how to solve this other than indeed separating the dev build setup to something like coq-universe.

Another solution would be to have two opam files, one for developer builds, and one for release builds. Indeed you can see that there are other problems if you diff the 8.16 branch opam file with master; opam handles vendoring not well at all.

Alizter:
OK I dropped touching the Dune version.

ejgallego:
Thanks!

PR #95, opened 2022-12-10
[fleche] Resumption of document checking

ejgallego:
We now resume document checking from the point it was left at. That
required some fixing upstream, see https://github.com/coq/coq/pull/16978 .

It seems to work well and be simple enough.

This does stress `Coq.State.compare`, so I did implement a much weaker
version; unfortunately we don't have a test-suite for checking that
memo is working properly, so we need to keep an eye on it and do some
manual testing.


ejgallego:
This seems to be working great! It makes a world of difference in large documents. Need to wait for Coq's PR to be merged tho, so I'm keeping it into draft status.

PR #98, opened 2022-12-16
README cleanup

artagnon:


ejgallego:
Thanks @artagnon ! Note that the reflow is due to emacs having `fill-column` at 70 by default, I'm happy with other values, what's the value chosen here?

ejgallego:
Ok thanks, will update the emacs config then.

ejgallego (approved):

PR #101, opened 2022-12-16
README: remove bogus instruction

artagnon:
It is not sfficient for `coq-lsp` to be in $PATH when developing.


ejgallego:
Indeed, thanks a lot! I added this because when we did the baby version of this server for lambdapi some developers preferred to do `dune install`.

This problem is a FAQ since Coq 8.16, as it requires `OCAMLPATH` to be setup properly to load plugins. A PITA.

ejgallego (approved):

PR #105, opened 2022-12-16
coq.tmLanguage: update from vscoq upstream, convert to json

artagnon:


editor/code/syntaxes/coq.json
```diff
@@ -0,0 +1,229 @@
+{
+    "fileTypes": ["v"],
+    "name": "Coq",
+    "patterns": [
+        {
+            "match": "\\b(From|Require|Import|Export|Local|Global|Include)\\b",
+            "comment": "Vernacular import keywords",
```
ejgallego, line 7: Technically `Global` and `Local` are import modifiers and not "Import Keywords" as `Require` or `Import` are; for example you have `Global Variable` etc...

editor/code/syntaxes/coq.json
```diff
@@ -0,0 +1,229 @@
+{
+    "fileTypes": ["v"],
```
ejgallego, line 2: Would be great if we could have some of this working for `.mv` files too, but I'm unsure how.

Note that we could also use LSP semantic highlighting.

ejgallego (approved):
LGTM, thanks!

ejgallego:
@artagnon can you add a changes entry?

PR #128, opened 2022-12-27
[infoview] Show Notice feedbacks on panel instead of as diagnostics

### LINKED ISSUE #125: [infoview] Show notices on panel instead of diagnostics
As of today, we show Coq's feedback `Notice` using diagnostics. This includes `Search`, `Check`, etc...

This is not good in most cases, instead, make that optional, and show them instead as messages in the goal / info panel.

ejgallego:
This seems more sensible as Coq (legacy) will produce large amounts of
these kind of Coq messages when for example `Search` or `About` is
used.

Thus, users can click on the concrete command and obtain the messages
information in the side panel; moreover, this allows for better
rendering opportunities.

We also implement a detailed display for errors, as Coq errors can be
large and don't fit properly in the diagnostics list.

A configuration option allows to restore old behavior.

Fixes #125


ejgallego:
Ready :)

PR #106, opened 2022-12-20
[check] Implement `fileProgress` notification for document progress

ejgallego:
This is modelled after Lean's `$/lean/fileProgress` notification: a
yellow bar in the right side panel indicates what parts of the
document are still not checked.

Fixes #54

Note that we send on fileProgress per sentence, that's likely too
much; for now, we throttle the decoration update. Could be made into a
user-config.

PR #129, opened 2022-12-28
fleche, editor: strip Ok_diagnostic feature

artagnon:
Since the introduction of #106, this feature is redundant, and only serves to choke up notifications in VSCode.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@
    (@ejgallego, #350)
  - Auto-ignore Coq object files; can be disabled in config
    (@ejgallego, #365)
+ - Remove the Ok diagnostics setting (@artagnon, #129)
 
 # coq-lsp 0.1.5.1: Path
 -----------------------
--- a/editor/code/package.json
+++ b/editor/code/package.json
@@ -112,11 +112,6 @@
             "default": true,
             "description": "Send diagnostics as document is processed; if false, diagnostics are only sent when Coq finishes processing the file"
           },
-          "coq-lsp.ok_diagnostics": {
-            "type": "boolean",
-            "default": false,
-            "description": "Send a diagnostic for sentences that are Ok; this can choke vsCode easily due to too many diagnostics"
-          },
           "coq-lsp.show_coq_info_messages": {
             "type": "boolean",
             "default": false,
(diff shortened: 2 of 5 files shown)


ejgallego (changes requested):
Nack, this is a very useful debug tool for me.

Alizter:
I am also unsure why this is considered redundant. Could you elaborate?

ejgallego:
I guess you can indeed determine when a sentence has been processed from the range sent in `fileProgress`.

We could also implement this in hover; so thinking a bit more about it, @artagnon is right about this functionality being a bit silly, however it is not clear if this functionality could be useful for other non-vscode clients.

I suggest we put this in the 0.2 milestone and discuss about it in the coq-lsp dev meeting that I plan to announce shortly.

ejgallego:
Moreover:
- this needs a changes entry
- while we have not been following semantic versioning, I hope that starting with 0.2 we do , so changes like this would have to wait for a minor release. Tho, due to some constraints client/server, our versioning scheme may be complex, at this stage it seems overkill to me to have to bump a minor version to change the goals request for example. Let's discuss about this.

Alizter:
I am OK with this in the end. I haven't used the Ok_diagnostic feature at all during debugging and I agree with @artagnon that it is no longer needed. Can you rebase?

ejgallego:
Now we have indeed the `fileProgress` notification which kind of serves the same purpose.

I'm still unsure if some other clients may want this, but I'm happy if you folks wanna get rid of this.

Alizter:
@ejgallego If clients want to have a "green highlighting feature" as we have discussed in the past, we should make it a custom LSP notification rather than hacking it into existing (unrelated) LSP notifications.

ejgallego:
> @ejgallego If clients want to have a "green highlighting feature" as we have discussed in the past, we should make it a custom LSP notification rather than hacking it into existing (unrelated) LSP notifications.

File progress works for that actually, what would be missing?

But note that the LSP protocol diagnostics are are only for errors and warnings, our use case is supported, it is just that the clients don't scale well it seems.

Alizter:
My thought would be to have a notification that simply emits a range in the document rather than a bunch of them. That would scale better.

ejgallego (approved):

PR #130, opened 2022-12-29
lsp: remove logfile, implement $/logTrace, $/setTrace

artagnon:
Instead of logging to log-lsp.txt, let us use $/logTrace to publish server-side logs to VSCode.

~~Unfortunately, it doesn't work, and I'm not able to figure out why: `trace` is always set to `off`. I'd appreciate any help.~~


controller/coq_lsp.ml
```diff
@@ -41,30 +41,13 @@ let option_default x d =
...
-    | "verbose" -> Verbose
-    | "off" -> Off
-    | _ -> raise (Invalid_argument "TraceValue.parse")
-
-  let to_string = function
-    | Off -> "off"
-    | Messages -> "messages"
-    | Verbose -> "verbose"
-end
+let ostring_field name dict = Option.map U.to_string (List.assoc_opt name dict)
```
Alizter, line 50: Why move this?

Alizter (approved):

controller/coq_lsp.ml
artagnon: ... because we need to read the value of `trace_value` in io.ml.

Alizter: Right but you moved it from line 44 to 51 I was curious why.

artagnon: Fixed now.

ejgallego:
Please set the tags and milestones properly before merging PRs, as well as changes when needed.

ejgallego (commented):
Thanks for this @artagnon ; however I think we may still need to be able to debug the protocol JSON itself from time to time.

How do we do this? I'm happy to use code for that, but if now, we should still have the log file for `Fleche.Debug.send` and `read`

lsp/io.ml
```diff
@@ -56,7 +77,6 @@ let mut = Mutex.create ()
 
 let send_json fmt obj =
   Mutex.lock mut;
-  if Fleche.Debug.send then Log.log_object "send" obj;
```
ejgallego, line 59: Is the option unused now?

controller/coq_lsp.ml
```diff
@@ -560,7 +546,7 @@ let lsp_main log_file std coqlib vo_load_path ml_include_path =
   let rec loop () =
     (* XXX: Implement a queue, compact *)
     let com = LIO.read_request stdin in
-    if Fleche.Debug.read then Log.log_object "read" com;
+    if Fleche.Debug.read then LIO.log_object "read" com;
```
ejgallego, line 549: Not sure this is useful in this form.

editor/code/src/client.ts
```diff
@@ -71,8 +71,6 @@ export function activate(context: ExtensionContext): void {
       if (goalPanel) goalPanel.dispose();
     }
 
-    window.showInformationMessage("Coq Language Server: starting");
```
ejgallego, line 74: Changes to this file should have gone in a separate PR, right?

How is this message now displayed to the user?

editor/code/src/client.ts
```diff
@@ -90,11 +88,14 @@ export function activate(context: ExtensionContext): void {
...
+    const serverOptions: ServerOptions = {
+      command: wsConfig.path,
+      args: wsConfig.args,
+    };
 
     client = new LanguageClient(
-      "coq-lsp-server",
-      "Coq Language Server",
+      "coq-lsp",
+      "Coq LSP Server",
```
ejgallego, line 98: Ditto, while I'm OK with this changes they seem unrelated to the PR title.

editor/code/src/client.ts
artagnon: This message is no longer displayed, since the trace displays a more reliable "server start" message.

editor/code/src/client.ts
artagnon: It's not unrelated: `coq-lsp-server.trace.server` was being read earlier, and this was a bug: `coq-lsp.trace.server` is set, and we read that.

lsp/io.ml
artagnon: No, but setting `coq-lsp.trace.server` to `verbose` supersedes this.

artagnon:
> Thanks for this @artagnon ; however I think we may still need to be able to debug the protocol JSON itself from time to time.
> 
> How do we do this? I'm happy to use code for that, but if now, we should still have the log file for `Fleche.Debug.send` and `read`

Set `coq-lsp.trace.server` to `verbose`.

artagnon:
Does this require a CHANGES entry?

lsp/io.ml
```diff
@@ -68,9 +88,31 @@ let logMessage fmt ~lvl ~message =
...
   send_json fmt msg
 
-let logTrace fmt ~message ?verbose () =
+let logTrace fmt ~verbose ~message =
   let method_ = "$/logTrace" in
-  let verbose = Option.cata (fun v -> [ ("verbose", `String v) ]) [] verbose in
-  let params = [ ("message", `String message) ] @ verbose in
-  let msg = Base.mk_notification ~method_ ~params in
-  send_json fmt msg
+  if !trace_value = TraceValue.parse verbose then
```
ejgallego, line 93: @artagnon I don't understand how this if is supposed to work, so `log_error` will produce no trace when `!trace_value = Verbose` ?

controller/coq_lsp.ml
ejgallego: Nope, the line should stay.

PR #131, opened 2022-12-30
ci: add format-check with prettier for editor/code

artagnon:
While at it, format all files with Prettier.


.github/workflows/build.yml
```diff
@@ -1,90 +1,106 @@
 name: Github CI
 
 on:
-  push:
-    branches:
-      - main
-  pull_request:
-    branches:
-      - main
+    push:
```
Alizter, line 4: Not a fan of the formatting in this file. I think you've got it set to 4 spaces per tab. I would prefer 2 spaces like before.

artagnon: @ejgallego said that he prefers 4 spaces, so we should be uniform in our formatting: we can't have 2 spaces for .yml files, and 4 spaces for .ts files.

artagnon: Actually, we format .ml files with 2 spaces, so I'm not sure why @ejgallego said that he prefers 4 spaces. I'll update prettier to use 2 spaces, and see if @ejgallego agrees with it.

Alizter: For ts files that is fine, but for yml, we quickly run out of room to read files on small screens.

Alizter:
Please don't format yml and md files with prettier. For ts do as you like.

artagnon:
I agree with your comment about .md files, but don't agree with you regarding .yml files: I've now added `.prettierignore` files to ignore .md files.

Alizter (approved):

Alizter:
For .ts, .js and .css it might be standard practice to have 4 spaces so I am not sure about those. I'll let @ejgallego decide.

ejgallego:
Yes, I think 4 spaces is better for C-like languages; note that I come from Linux coding style tradition which I take the freedom to [quote](https://www.kernel.org/doc/html/v4.11/process/coding-style.html) here:

> Tabs are 8 characters, and thus indentations are also 8 characters. There are heretic movements that try to make indentations 4 (or even 2!) characters deep, and that is akin to trying to define the value of PI to be 3.

So while I do prefer 8 spaces, indeed I understand that's too much for some devs, so I find 4 a good compromise.

OCaml and Haskell are a different story as they don't really have control blocks other than patter matching which is nested.

ejgallego:
Also note this:

> Now, some people will claim that having 8-character indentations makes the code move too far to the right, and makes it hard to read on a 80-character terminal screen. The answer to that is that if you need more than 3 levels of indentation, you’re screwed anyway, and should fix your program.

PR #135, opened 2022-12-31
Bump deps

artagnon:


ejgallego (changes requested):
What are these sporadic failures?

Nack on the 4.14 part, 0.1.0 had a lower bound of 4.13 and that was a problem but quite a lot of users which are still bound to lower versions.

In general, I'm not a fan of restricting deps artificially, `coq-lsp` should aim to support all OCaml versions Coq supports unless there is a reason.

artagnon:
> What are these sporadic failures?
> 
> Nack on the 4.14 part, 0.1.0 had a lower bound of 4.13 and that was a problem but quite a lot of users which are still bound to lower versions.
> 
> In general, I'm not a fan of restricting deps artificially, `coq-lsp` should aim to support all OCaml versions Coq supports unless there is a reason.

Fixed now. The sporadic failures were seen on macos, but it didn't look like my patch helped in any way: looks like it was due to changes to macos runners on Github; anyway, it's fixed now.

.github/workflows/build.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         run: opam exec -- make opam-deps
 
       - name: 🧱 Build coq-lsp
-        run: opam exec -- make build
+        run: opam exec -- make
```
ejgallego, line 49: Why this change? Relying on the default target seems fragile to me.

coq-lsp.opam
```diff
@@ -10,34 +10,38 @@ homepage: "https://github.com/ejgallego/coq-lsp"
...
+  "Shachar Itzhaky <shachari@cs.technion.ac.il>"
+  "Ramkumar Ramachandra <r@artagnon.com>"
 ]
 license: "LGPL-2.1"
 doc: "https://ejgallego.github.io/coq-lsp/"
 
 depends: [
   "ocaml"        {         >= "4.12.0" }
-  "dune"         { build & >= "3.5.0"  }
+  "dune"         { build & >= "3.6.1"  }
```
ejgallego, line 22: Why this change, in fact our minimal supported dune is 3.2, we use 3.5 only because of the vendoring.

coq-lsp.opam
```diff
@@ -10,34 +10,38 @@ homepage: "https://github.com/ejgallego/coq-lsp"
...
-  "ppx_hash"            {           >= "v0.13.0"  }
-  "ppx_import"          { build   & >= "1.5-3"    }
-  "ppx_deriving_yojson" {           >= "3.4"      }
+  "sexplib"             {           >= "v0.15.1"  }
+  "ppx_deriving"        {           >= "5.2.1"    }
+  "ppx_sexp_conv"       {           >= "v0.15.1"  }
+  "ppx_compare"         {           >= "v0.15.0"  }
+  "ppx_hash"            {           >= "v0.15.0"  }
+  "ppx_import"          { build   & >= "1.10.0"   }
+  "ppx_deriving_yojson" {           >= "3.7.0"    }
```
ejgallego, line 38: This is too restrictive for some users on older OCaml versions, they explicitly demanded we support them and a big users at Inria, so nack.

coq-lsp.opam
```diff
@@ -10,34 +10,38 @@ homepage: "https://github.com/ejgallego/coq-lsp"
...
-  "yojson"       { >= "1.7.0" }
+  "cmdliner"     { >= "1.1.1" }
+  "yojson"       { >= "2.0.2" }
 ]
 
-build: [ [ "dune" "build" "-p" name "-j" jobs ] ]
-run-test: [ [ "dune" "runtest" "-p" name "-j" jobs ] ]
+build: [
+  [ "dune" "build" ]
+]
```
ejgallego, line 47: This is wrong.

editor/code/package.json
```diff
@@ -144,12 +144,12 @@
     ]
   },
   "devDependencies": {
-    "@types/node": "^10.14.15",
+    "@types/node": "^18.11.18",
```
ejgallego, line 147: For now our minimal node version is 16, so we can bump to 16.

editor/code/package.json
```diff
@@ -144,12 +144,12 @@
...
     "@types/throttle-debounce": "^5.0.0",
-    "@types/vscode": "^1.73.0",
-    "esbuild": "^0.15.14",
+    "@types/vscode": "^1.74.0",
```
ejgallego, line 149: For now our base is 1.73, do we need 1.74 for some feature? Otherwise let's stay a 1.73.

Rest of changes are ok to me.

artagnon:
Closing, as it's not worth re-rolling.

PR #138, opened 2023-01-02
vendor Coq-HoTT to demonstrate bugs

Alizter:
Vendoring Coq-HoTT and building doesn't seem to work with coq-lsp. I observe the following lacking behaviours:

- -noinit is not being passed
- -R/-Q flags are not being passed

This is probably something about _CoqProject not being read correctly.

Also I should note that Coq-HoTT is a Dune project, so ideally we should also be able to read Dune files too.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -6,3 +6,6 @@
 	path = vendor/coq-serapi
 	url = https://github.com/ejgallego/coq-serapi.git
 	branch = v8.17
+[submodule "Coq-HoTT"]
+	path = Coq-HoTT
+	url = https://github.com/HoTT/Coq-HoTT.git


artagnon:
Nack, `_CoqProject` is read from the root of the workspace, so a submodule won't help.

Alizter:
Shouldn't the closest parent _CoqProject to the file be chosen? Also I don't know what workspace means in this situation. I don't think it means the dune meaning right?

artagnon:
I suppose, but VSCoq only supports `_CoqProject` at the root of the workspace as well. You could open `coq-lsp/Coq-HoTT` in VSCode and try it. Besides, there is a way to tell dune to use a different root dir: I think you can simply clone Coq-HoTT and execute `dune exec --root=<path to coq-lsp> -- dune build`; I'm not 100% sure about the exact command though.

ejgallego:
For now indeed we only support single root workspace, but we should eventually support workspace folders.

I think tho that workspace folders makes more sense once our single folder workspace support is mature, so it looks to me like something we would like to ship in 0.3 or so.

PR #140, opened 2023-01-02
workspace: add parsing for -noinit, -indices-matter, -impredicative-set

artagnon:


ejgallego (approved):
Great thanks!

See comment if you feel like improving it.

ejgallego:
@artagnon this deserves IMHO a changes entry.

coq/workspace.ml
```diff
@@ -19,8 +19,7 @@ module Setup = struct
   type t =
     { vo_load_path : Loadpath.vo_path list
     ; ml_include_path : string list
-    ; options : (Goptions.option_name * Coqargs.option_command) list
```
ejgallego, line 22: @artagnon actually I used this type as to provide a finer structure in terms of what command line arguments are passed. For example we could have field records for warnings, for kernel options, etc...

No problem refining that later.

artagnon: `Goptions.option_name * Coqargs.option_command` is not appropriate for non-Vernac options.

Alizter: We can have something more sophisticated like a variant, but as Emilio says no need at the moment.

ejgallego: Yes, it is not enough, what I mean is that we should at some point using something a bit finer than string.

`Goptions.option_name` worked fine both for warnings and options tho.

ejgallego:
Thanks, don't forget to set the labels before merging !

ejgallego:
Does this fix #138 ?

artagnon:
Partially. Many files in Coq-HoTT work, but several bugs remain.

Alizter:
I am getting a lot of anomalies when running in Coq-HoTT. I need `--bt` to find out where they are coming from though.

ejgallego:
@Alizter set `Debug.backtraces` to true and re compile.

fleche/doc.ml
```diff
@@ -44,8 +44,14 @@ type t =
 
 let mk_doc root_state workspace =
   let libname = Names.(DirPath.make [ Id.of_string "foo" ]) in
-  let require_libs = [ ("Coq.Init.Prelude", None, Some (Lib.Import, None)) ] in
-  Coq.Init.doc_init ~root_state ~workspace ~libname ~require_libs
+  if !Coq.Workspace.no_init = true then
```
ejgallego, line 47: @Alizter @artagnon this is bogus as `Workspace.apply` is called in `doc_init` below. That likely makes your trouble for HoTT.

ejgallego: The right fix is to put require_libs in the workspace setup.

Issue #96, opened 2022-12-16
[goal printing] Use jsCoq goal printing code


ejgallego:
Current goal printing is very basic and doesn't look good at all; in particular we didn't set indentation and margins on, also there is no coloring.

Best way to fix this is to use jsCoq's `Pp.t -> Html` routine (https://github.com/jscoq/jscoq/blob/2a60fd083b6b09cbe19f750f31a877740d140ea2/frontend/classic/js/format-pprint.js) , which is well-tested and provides some nice features.

ejgallego:
Fixed by #143

PR #143, opened 2023-01-03
goals: lay foundation for pretty-printing goals (#96)

artagnon:
Integrate pretty-printer from jsCoq to pretty-print goals.


coq/goals.mli
```diff
@@ -40,6 +40,24 @@ type 'a goals =
...
+  type t =
+    | Pp_empty
+    | Pp_string of string
+    | Pp_glue of t list
+    | Pp_box of block_type * t
+    | Pp_tag of pp_tag * t
+    | Pp_print_break of int * int
+    | Pp_force_newline
+    | Pp_comment of string list
+  [@@deriving yojson]
```
ejgallego, line 56: SerAPI already provides the yojson for `Pp.t` , indeed the seralization should happen in the lsp layer not here

editor/code/src/goals.ts
```diff
@@ -27,17 +47,153 @@ interface GoalAnswer {
...
+    case BreakMode.Horizontal:
+      return "horizontal";
+    case BreakMode.Vertical:
+      return "vertical";
+    case BreakMode.SkipVertical:
+      return "skip-vertical";
+  }
+}
+
+function pp2HTML(msg: Pp, state: PpState): string {
```
ejgallego, line 61: Let's put this in its own file.

coq/goals.mli
artagnon: I thought `coq-lsp` would subsume `serapi` ... I read it somewhere in the README.

ejgallego:
Thanks @artagnon , I think extra css stuff is at 

- https://github.com/jscoq/jscoq/blob/v8.16/frontend/classic/css/coq-base.css
- https://github.com/jscoq/jscoq/blob/v8.16/frontend/classic/css/coq-dark.css
- https://github.com/jscoq/jscoq/blob/v8.16/frontend/classic/css/coq-light.css

Sadly the relevant parts are not extracted.

The goal is to have a common lib for this, so let's try to keep the files one to one as much as we can with jsCoq, including submitting a patch to jsCoq.

editor/code/src/goals.ts
artagnon: Done.

artagnon:
I've copied out the relevant part of the css, but unfortunately, js is also required: see `frontend/classic/js/format-pprint.js`.

coq/goals.mli
ejgallego: Yes, it will subsume serapi, not serlib tho, sorry for the confusion.

editor/code/src/types.ts
```diff
@@ -0,0 +1,48 @@
...
+  ty: EntryType;
+  hyps: Hyp[];
+}
+
+export interface GoalRequest {
+  textDocument: TextDocumentIdentifier;
+  position: Position;
+}
+
+export interface GoalAnswer {
```
ejgallego, line 44: Why is this a better place than in `goals.ts` where the users are?

coq/goals.mli
ejgallego: Also note that for now we have kept the `coq` library free of deps, so it could be merged in upstream at some point.

editor/code/src/types.ts
artagnon: It's normally good practice to keep all the types in one place, for easy reference: think of it as protocol documentation.

coq/goals.mli
artagnon: Fixed.

ejgallego:
Taking the lock to rebase and do some tweaks.

ejgallego:
> The relevant part of the css is all the style for `div.Pp_box` and `span.Pp_break`: I've copied those out, but unfortunately, it doesn't seem to work as intended.

The styles missing are in the files I linked, `constr` `tactic` etc....

artagnon:
> > The relevant part of the css is all the style for `div.Pp_box` and `span.Pp_break`: I've copied those out, but unfortunately, it doesn't seem to work as intended.
> 
> The styles missing are in the files I linked, `constr` `tactic` etc....

Thanks! I've added them now. However, what I'm talking about is line breaks and indents: they don't seem to be working as expected.

ejgallego:
> Thanks! I've added them now. However, what I'm talking about is line breaks and indents: they don't seem to be working as expected.

I saw that too, likely some missing bits on CSS, I'm having a look.

The rest looks pretty good thanks, I'm gonna do a bit of reorg both in jsCoq and here so we can share the code more easily, and merge.

ejgallego:
That gives a hint on what the problem may be:
```coq
Definition m := addn0.
Print m.
```
```
m = addn0Pp_force_newline : forall n : nat, n + 0 = nPp_force_newlinePp_force_newlineArguments m n%nat_scope
```

ejgallego:
That works correctly in jsCoq so we should be able to debug it.

I guess it is maybe better not to diverge too much from upstream (jsCoq) for now, as to debug this kind of issues.

ejgallego:
Actually it seems to me that you did port the old code from jsCoq `format-pprint.js` right?

artagnon:
Yes, that's right. I don't want to depend on jQuery, so `pp2DOM` is probably a non-starter.

artagnon:
jQuery is an old and slow library written in js. I'd much rather depend on cash if we were to go down that route, as it's a typescript library, but we have to adapt quite a bit. In unrelated comments, please clean up the PR: there are conflict markers in my commit.

ejgallego:
Using cash sounds good, even better if we use plain DOM for this small code. jsCoq already was using jQuery so there is not deep reason the code used it.

ejgallego:
We are in big trouble here, the jsCoq renderer has to run in the webview context.

I suggest we proceed as follows:
- implement the webview goal rendering message, move the current rendering in `main` to do this (in a separate PR)
- adapt this PR to that protocol.

ejgallego:
I'm postponing the milestone as this will require deep changes in the infoview.

artagnon:
Okay, this branch is ready to review now.

Alizter:
However there are still some issues with rendering newlines. See T2048.v:
image

artagnon:
Can you try it on jsCoq? If jsCoq doesn't render it correctly, we have to fix the pretty-printer there.

Alizter:
This is how it renders in CoqIDE FTR:
image

Alizter:
And in jscoq:
image

artagnon:
Hm, that's unfortunate. We're probably not calling some essential function. In any case, I think we should merge this, and tackle it in a follow-up patch, especially since coq-lsp crashes on T2048.v file for me.

Alizter:
Interestingly when displaying the goal window, the behavior is different:
```coq
Require Import Ascii String List.

Definition nl := (String (ascii_of_nat 10) "")%string.

Import ListNotations.
Local Open Scope string.

Definition paragraph := ["These"; "Should"; "Be"; "On"; "New"; "Lines"].

Definition para_string := String.concat nl paragraph.

Axiom P : string -> Prop.

Goal P para_string.
cbv.
```
LSP:
image
CoqIDE:
image
jscoq:
image

ejgallego:
Thanks for all the tests Ali, there are very useful.

Alizter:
The goal window doesn't update before a non-qed error: https://github.com/ejgallego/coq-lsp/issues/191

ejgallego (commented):
The PR is IMHO in great shape, however the rendering doesn't seem to be working 100% fine yet.

I'm going to investigate more.

Some more stuff that we may need to do (in this PR or later will depend on well we can get this PR working without it, IMHO):
- allow the server to send `Pp | string`
- retrieve the code for bullet printing etc... from jsCoq.

ejgallego:
Ok folks, I did some more investigations. The main problem we have with the current code is that it is very hard for us to adjust the rendering where we should with the current panel implementation.

I propose then we get ahead with #223 , then fine tuning this PR will be **much, much** easier, thanks to `useLayoutEffect` in React. Moreover a lot of other nice things are just very implement on top of that.

I will take care of the rebase.

ejgallego:
Ok folks, this PR is now almost ported to React; it draws but need to read a bit more about how to do some stuff, in particular clean up the mess with the jquery integration.

For now this seems to bug as it doesn't get the right width for the element, mmmm, need to see what's going on.

By the way I think it makes sense to port `pp2DOM` to react.

Alizter:
Hmm
image

Alizter:
Regression on the new lines being printed (they work in master now):
image

artagnon:
Okay folks, after much debugging, I'm pleased to report that the goal-printing code is now fixed! Modulo a rebase (for @ejgallego), this branch is ready.

Alizter:
Error messages are incorrectly formatted.
image

ejgallego:
@Alizter looks like it, the width of the parent container doesn't get propagated.

Alizter:
Error messages are now fixed.

Alizter:
```coq
Goal nat.
pose (x := 2 + 4 + 5).
```
produces:
image

Which is a bit incorrectly formatted.

Alizter:
Works now.

ejgallego:
Note that we have a vim mode already, so we need to put an option in the init as to choose between string and Pp.t, as otherwise we'd break the vim plugin.

PR #141, opened 2023-01-03
[controller] Allow to postpone and cancel requests

ejgallego:
Some requests need to be delayed until the document is ready, such as
`documentSymbols`. In order to do this, we allow to postpone requests,
and `Doc_manager.check` will now return the list of requests pending
for a given document.

The implementation is lightweight and quite a prototype, but should
serve as a basis for more advanced stuff.

Closes #124.

PR #145, opened 2023-01-03
Clean up some code, improve debug info in fleche

artagnon:


ejgallego (changes requested):
Instead of duplicate the `find` check at every place the approach in #141 is preferred.

ejgallego:
Ok, you can rebase on top of #146

fleche/doc.ml
```diff
@@ -221,7 +218,7 @@ let state_recovery_heuristic doc st v =
   | _ -> st
 
 let debug_parsed_sentence ~ast =
-  let loc = Coq.Ast.loc ast |> Option.get in
+  let loc = Coq.Ast.loc ast |> Option.default (Loc.make_loc (0, 0)) in
```
ejgallego, line 221: This is incorrect.

fleche/memo.ml
```diff
@@ -144,7 +144,7 @@ let interp_command ~st ~fb_queue stm : _ Stats.t =
       fb_queue := [];
       Stats.make ~time res
     | Coq.Protect.R.Completed _ as res ->
-      let () = HC.add !cache (stm, st) (stm_loc, res) in
+      HC.add !cache (stm, st) (stm_loc, res);
```
ejgallego, line 147: That's using the let for a reason (robusnest of syntax)

Let's not spend time doing these pointless changes.

fleche/doc.ml
artagnon: What do you mean? If `Coq.Ast.loc` returns `None` loc, what do we do?

fleche/doc.ml
```diff
@@ -162,9 +163,6 @@ let report_progress ~doc last_tok =
   let progress = compute_progress doc.end_loc last_tok in
   Io.Report.fileProgress ~uri:doc.uri ~version:doc.version [ progress ]
 
-(* XXX: Save on close? *)
-(* let close_doc _modname = () *)
-
```
ejgallego, line 167: The comments are placeholders for a reason.

fleche/doc.ml
ejgallego: As all the other changes, if the location is not there an error must be thrown.

I know this is not perfect but let's keep the Option.get for now

fleche/doc.ml
```diff
@@ -292,7 +289,6 @@ let process_and_parse ~uri ~version ~fb_queue doc last_tok doc_handle =
   let rec stm doc st last_tok =
     let doc = send_eager_diagnostics ~uri ~version ~doc in
     report_progress ~doc last_tok;
-    if Debug.parsing then Io.Log.error "coq" "parsing sentence";
```
ejgallego, line 295: Why this is gone?

artagnon: There's another line showing the parsed sentence, which I find more useful.

fleche/doc.ml
ejgallego: It will never return `None`, the typing upstream is too weak due to the ast type being used for both parsing and term synthesis.

fleche/doc.ml
ejgallego: There are different concepts, this line is here to let us know parsing has started, and the server is not stuck in some other place.

ejgallego: As a general rule I'd suggest we discuss first any further change to the logging infrastructure. I have pretty specific logging needs that can only be discovered if you work in the core engine and all these changes are not suiting my workflow well.

artagnon:
All fixed.

ejgallego (approved):
Thanks, rebased on top of #148 , the extra verbosity of `print_file:true` didn't work well in my setup.

PR #147, opened 2023-01-03
feature: add --bt flag

Alizter:
I've removed it from the debug.ml file since it makes it awkward to turn on from there.

### DIFF
--- a/controller/coq_lsp.ml
+++ b/controller/coq_lsp.ml
@@ -532,7 +532,7 @@ let log_workspace (_, from) =
   let message = "Configuration loaded from " ^ from in
   LIO.log_info "config" message
 
-let lsp_main std coqlib vo_load_path ml_include_path =
+let lsp_main bt std coqlib vo_load_path ml_include_path =
   LSP.std_protocol := std;
   Exninfo.record_backtrace true;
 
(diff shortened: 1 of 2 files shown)


controller/coq_lsp.ml
```diff
@@ -545,7 +545,7 @@ let lsp_main std coqlib vo_load_path ml_include_path =
 
   (* Core Coq initialization *)
   let fb_handler, fb_queue = mk_fb_handler () in
-  let debug = Fleche.Debug.backtraces in
+  let debug = bt in
```
ejgallego, line 548: Why not `bt || Fleche.Debug.backtraces` ?

ejgallego (commented):
LGTM, tho I still find the settings in `debug.ml` faster than going to vscode and editing a config setup, so let's keep both.

ejgallego (approved):
Thanks Ali!

ejgallego:
Actually @Alizter I think the most useful setting would be to allow to set this from the extension settings in vscode.

Alizter:
@ejgallego That is already possible using the flags field. We could add an option that does this too. Though I am not a big fan of multiple ways of doing the same thing.

ejgallego:
It seems to me that command line flags are not very discoverable in this situation, indeed clients should pass options to the server using the protocol as code does now.

So I'd remove the cmdline flag and leave only the option then.

The fact that `coq-lsp` supports Coq's command line flags is mostly a historical accident, configuration of the workspace should always happen via _CoqProject of Dune IMHO.

PR #148, opened 2023-01-04
[logging] Several tweaks to #130

ejgallego:
See individual commits for details.


lsp/io.ml
```diff
@@ -74,9 +74,11 @@ let read_request ic =
   | Invalid_argument msg -> raise (ReadError msg)
 
 let mut = Mutex.create ()
+let log = ref (fun _ _ -> ())
```
artagnon, line 79: I'm not too fond of this hack: can you try `let ... and let` instead?

controller/coq_lsp.ml
```diff
@@ -501,15 +500,15 @@ let rec process_queue ofmt ~state =
...
+    LIO.trace "process_queue" ("Serving Request: " ^ m);
     (* We let Coq work normally now *)
     Control.interrupt := false;
     dispatch_message ofmt ~state dict);
   process_queue ofmt ~state
 
 let lsp_cb oc =
   Fleche.Io.CallBack.
-    { log_error = LIO.log_error
+    { log_error = LIO.trace ?extra:None
```
artagnon, line 511: Maybe rename this field to `trace_verbose`, and perhaps introduce a `trace_message`?

lsp/io.ml
ejgallego: Indeed that's a good idea, but I have another PR to improve some of the limitations of `trace` logging where this goes away.

controller/coq_lsp.ml
ejgallego: Renamed, I think just `trace` + extra info for verbose is ok.

PR #162, opened 2023-01-09
[server] add version string in separate file and update version, fix bug

Alizter:
This should make it easier to update the version when doing a release.

### DIFF
--- a/controller/coq_lsp.ml
+++ b/controller/coq_lsp.ml
@@ -74,12 +74,15 @@ let do_client_options coq_lsp_options =
   | Ok v -> Fleche.Config.v := v
   | Error msg -> LIO.trace "CoqLspOption.of_yojson error: " msg
 
-let check_client_version version : unit =
-  LIO.trace "version" version;
-  match version with
-  | "any" | "0.1.3" -> ()
-  | v ->
-    let message = Format.asprintf "Incorrect version: %s , expected 0.1.2" v in
+let check_client_version client_version : unit =
+  LIO.trace "client_version" client_version;
+  if String.(equal client_version "any" || equal client_version Version.server)
+  then () (* Version OK *)
+  else
+    let message =
+      Format.asprintf "Incorrect client version: %s , expected %s."
+        client_version Version.server
+    in
     LIO.logMessage ~lvl:1 ~message
 
 let do_initialize ofmt ~id params =
@@ -704,7 +707,7 @@ let lsp_cmd : unit Cmd.t =
   let vo_load_path = term_append [ rload_path; load_path ] in
   Cmd.(
     v
-      (Cmd.info "coq-lsp" ~version:"0.01" ~doc ~man)
+      (Cmd.info "coq-lsp" ~version:Version.server ~doc ~man)
       Term.(const lsp_main $ bt $ std $ coqlib $ vo_load_path $ ml_include_path))
 
 let main () =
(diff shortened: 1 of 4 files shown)


controller/version.mli
```diff
@@ -0,0 +1,19 @@
...
+(*    //   *    This file is distributed under the terms of the         *)
+(*         *     GNU Lesser General Public License Version 2.1          *)
+(*         *     (see LICENSE file for the text of the license)         *)
+(************************************************************************)
+
+(************************************************************************)
+(* Coq Language Server Protocol                                         *)
+(* Copyright 2019 MINES ParisTech -- Dual License LGPL 2.1 / GPL3+      *)
+(* Copyright 2019-2022 Inria      -- Dual License LGPL 2.1 / GPL3+      *)
+(* Written by: Emilio J. Gallego Arias                                  *)
```
ejgallego, line 15: You can remove the second part of this header I think.

ejgallego (changes requested):
Looks good to me, but:
- version is 0.1.3
- you need to update check_client_version too

Alizter:
@ejgallego I've updated check_client_version to use this now. I have also made the version API richer.

controller/version.ml
```diff
@@ -0,0 +1,71 @@
...
+  | Any, Any -> `Equal
+  | Any, _ -> `More
+  | _, Any -> `Less
+  | Version a, Version b -> (
+    match compare_int a.major b.major with
+    | `Equal -> (
+      match compare_int a.minor b.minor with
+      | `Equal -> compare_int a.patch b.patch
+      | x -> x)
+    | x -> x)
```
ejgallego, line 39: That seems pretty over engineered to me, moreover, I don't want to maintain this code here, I'd rather use `semver2` if we ever have a need.

Let's keep things simple for now.

controller/version.mli
```diff
@@ -0,0 +1,30 @@
...
+val to_string : t -> string
+
+(** Parse a string to a version. *)
+val of_string : string -> t option
+
+(** Parse a string to a version and throw an exception if invalid. *)
+val of_string_exn : string -> t
+
+(** Compare two versions. *)
+val compare : t -> t -> [ `Equal | `Less | `More ]
```
ejgallego, line 30: Please don't use variants for use cases where they are not required (as discussed in the Dune context)

controller/coq_lsp.ml
```diff
@@ -76,10 +76,12 @@ let do_client_options coq_lsp_options =
 
 let check_client_version version : unit =
   LIO.trace "version" version;
-  match version with
-  | "any" | "0.1.3" -> ()
-  | v ->
-    let message = Format.asprintf "Incorrect version: %s , expected 0.1.2" v in
+  if Version.equal (Version.of_string version) Version.current then ()
```
ejgallego, line 79: PR looks good to me thanks, tho I'm unsure if this whole dance `of_ / to_` is worth it.

If I was reading this, I'd be confused about the semantics of `Version.equal`, IMVHO older code was more explicit.

ejgallego (commented):
Another point is that maybe the version file should be auto-generated (from `package.json`?) but that leads to modularity issues.

Ummm, proper setup is too complex for us, because it would consist of 3 packages: `protocol` / `client` / `server`

ejgallego (approved):

PR #181, opened 2023-01-13
[code] [protocol] Add notification about document perf data.

ejgallego:
We will include several interesting information, such as a table with memory and time consumption, etc...

Waiting on the webview build infrastructure to finish it.


Alizter (commented):
Looks OK, still need to try it out tho.

ejgallego:
UI is very preliminary, that's really a way to have a window for us to put random stuff and generally do experiments.

PR #188, opened 2023-01-18
editor/code: scope keybinding alt+enter correctly

artagnon:
The keybinding alt+enter is also used in find-and-replace, when the user wants to replace all instances. Before this change, the keybinding supplied by coq-lsp used to interfere. Moreover, the keybinding should be scoped for just the coq language.


ejgallego (approved):
Thanks a lot Ram!

PR #198, opened 2023-01-19
editor/code/test: lay the foundation for writing tests

artagnon:
We have picked the Jasmine testing framework, and included a test for extension activation. Unfortunately, there seems to be some misunderstanding with Jasmine Spies, as they don't seem to capture the activation information message.


.gitignore
```diff
@@ -2,12 +2,6 @@ _build
...
 *.install
-log-*.txt
-
 .git-ps
 
 *.cache
-
-editor/code/out
-
-result
```
Alizter, line 13: This gets produced by `nix build`.

artagnon: This belongs to the commit in #171. Only the last commit is the subject of this PR.

Alizter:
It would be good to resurrect this. @artagnon could you list what is left to be done other than rebase?

artagnon:
I think this is the wrong approach, and that we should pick `wdio-vscode` as the testing framework.

artagnon:
Actually, I'm not sure anymore. @ejgallego pointed me to https://github.com/coq-community/vscoq/pull/372, which is where I got the idea for `wdio-vscode`. Unfortunately, that PR seems to be closed, citing reasons of flakiness.

Writing tests independently of vscode is another alternative, but after some digging, it seems that the only full-featured LSP clients are those that ship with editors; there seems to be no real LSP client written in Typescript independent of VSCode APIs, using the client portion of ocaml-lsp-server will blow up the size of our tests, including lots of irrelevant details like YoJson parsing. I will continue investigating.

PR #200, opened 2023-01-19
[coq] [fleche] Unicode Support

### LINKED ISSUE #193: Error locations get skewed in the presence of unicode characters
This is likely an upstream bug, but when in the presence of unicode characters, the reported error locations are skewed:
image

You can try this example for yourself:
```coq
Definition foo : forall Γ Γ' (Γ ⊆ Γ') : Type.
```
It isn't just parser errors, this is just a minimal example.

### LINKED ISSUE #197: Goal not updating after `intros`
Here is an example of the goal not updating after an `intros` statement together with some unicode characters.

```coq
Axiom P : nat -> Prop.
Axiom foo : forall {n}, P (S n) -> P n.
Ltac foo := apply foo.
Infix "⊆" := lt (at level 10).

Goal forall Γ Δ, Γ ⊆ Δ -> P Γ.
intros Γ Δ s.
foo.
```
Cursor is first before the `intros`.
image

Cursor is after the terminating `.` and the goal has not updated.
image

Next line the goal updates.
image

Notice the correct behaviour when the cursor is after the terminating `.` for `foo`.
image

ejgallego:
We enforce a stricter separation between `Range.t` and `Loc.t`, with
Flèche not using `Loc.t` anymore, and `Range.t` being in unicode
character positions.

To achieve this we need to convert all the positions coming from Coq;
we do this in a principled way, but the whole business is often
delicate.

As we keep a position per sentence, this implies a `O(n)` number of
conversions in total, where `n is the number of sentences; we could
optimize that to be lazy, but it is unlikely to show in profiles as
that is a pretty reasonably `O(.)` factor.

Fixes #193 , fixes #197 .

### DIFF
--- a/fleche/debug.ml
+++ b/fleche/debug.ml
@@ -10,6 +10,7 @@ let cache = false || all
 (* LSP messages: Send and receive *)
 let send = false || all || lsp
 let read = false || all || lsp
+let lsp_init = false || all || lsp
 
 (* Parsing (this is a bit expensive as it will call the printer *)
 let parsing = false || all
@@ -19,6 +20,9 @@ let scan = false || all
 
 (* Backtraces *)
 let backtraces = false || all
+
+(* Unicode conversion *)
+let unicode = false || all
 
 (* Sched wakeup *)
 let sched_wakeup = false || all
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,9 +1,10 @@
-# coq-lsp 0.1.4:
-----------------------
+# coq-lsp 0.1.4: View
+---------------------
 
  - Support for OCaml 4.11 (@ejgallego, #184)
  - The keybinding alt+enter in VSCode is now correctly scoped
    (@artagnon, #188)
+ - Support Unicode files (@ejgallego, #200, fixes #193, fixes #197)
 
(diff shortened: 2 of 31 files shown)


fleche/dune
```diff
@@ -1,4 +1,4 @@
 (library
  (name fleche)
  (public_name coq-lsp.fleche)
- (libraries coq))
+ (libraries coq camomile))
```
ejgallego, line 4: We don't need to depend on camomile, actually what we need is so simple that we could maybe just vendor it, should be very few lines of code (actually Coq already has the required UTF-8 functions but they are not exporting them)

fleche/debug.ml
```diff
@@ -18,7 +19,10 @@ let parsing = false || all
 let scan = false || all
 
 (* Backtraces *)
-let backtraces = false || all
+let backtraces = false || all || true
```
Alizter, line 22:
```suggestion
let backtraces = false || all
```

Alizter (approved):

ejgallego:
@Alizter I think I'm happy with the code now; how did your testing go?

Alizter:
@ejgallego my issues have been resolved.

PR #208, opened 2023-01-21
[lsp] Remove noop --std argument.

ejgallego:
This argument was introduced very long time ago in the Lambdapi implementation as we were not fully LSP compliant, but now we are.

The argument itself has been disabled since a long time, and it misleading now; moreover we put it as default in `package.json`!!!

### DIFF
--- a/controller/coq_lsp.ml
+++ b/controller/coq_lsp.ml
@@ -141,10 +141,10 @@ let do_initialize ~params =
 let rtable : (int, PendingRequest.t) Hashtbl.t = Hashtbl.create 673
 
 let answer_request ~ofmt ~id result =
-  match result with
-  | Result.Ok result -> LSP.mk_reply ~id ~result |> LIO.send_json ofmt
-  | Error (code, message) ->
-    LSP.mk_request_error ~id ~code ~message |> LIO.send_json ofmt
+  (match result with
+  | Result.Ok result -> LSP.mk_reply ~id ~result
+  | Error (code, message) -> LSP.mk_request_error ~id ~code ~message)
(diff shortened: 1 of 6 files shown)


controller/coq_lsp.ml
```diff
@@ -568,7 +569,7 @@ let lsp_main bt std coqlib vo_load_path ml_include_path =
 (* Arguments handling *)
 open Cmdliner
 
-let std =
+let _std =
```
Alizter, line 572: Why not completely remove?

ejgallego: > Why not completely remove?

Yes, let's get rid of it.

ejgallego:
For a historical note, we needed this because in the first versions of the Lambdapi prototype, we sent the goals inside the diagnostics; this was not standard- compliant. We have now moved from that paradigm, and `coq-lsp` is fully standard compliant now TTBOMK.

If clients don't issue non-standard requests, they won't see any non-standard messages except for the `$/` notifications, but these are legal by https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#dollarRequests:

> Notification and requests whose methods start with ‘$/’ are messages which are protocol implementation dependent and might not be implementable in all clients or servers. For example if the server implementation uses a single threaded synchronous programming language then there is little a server can do to react to a `$/cancelRequest` notification. If a server or client receives notifications starting with ‘$/’ **it is free to ignore the notification**.

Alizter:
Maybe getting rid of all the command line arguments will be for the better.

Alizter (approved):

PR #210, opened 2023-01-22
[code] Check for VSCoq and complain about a possible conflict.

### LINKED ISSUE #183: [code] We need to manage when both `coq-lsp` and VsCoq are installed.

ejgallego:
I'm unsure we can do better.

Closes #183


Alizter:
Seems ok to me.

PR #212, opened 2023-01-22
categorize settings

Alizter:
We group settings into related categories.

### DIFF
--- a/editor/code/package.json
+++ b/editor/code/package.json
@@ -75,8 +75,62 @@
     ],
     "configuration": [
       {
+        "title": "Diagnostics",
         "type": "object",
-        "title": "Coq LSP configuration",
+        "properties": {
+          "coq-lsp.eager_diagnostics": {
+            "type": "boolean",
+            "default": true,
+            "description": "Send diagnostics as document is processed; if false, diagnostics are only sent when Coq finishes processing the file"
+          },
+          "coq-lsp.ok_diagnostics": {
+            "type": "boolean",
+            "default": false,
+            "description": "Send a diagnostic for sentences that are Ok; this can choke vsCode easily due to too many diagnostics"
+          },
+          "coq-lsp.show_coq_info_messages": {
+            "type": "boolean",
+            "default": false,
+            "description": "Show Coq's \"Info\" messages as diagnostics, such as 'foo has been defined.' and miscellaneous operational messages."
+          },
+          "coq-lsp.show_notices_as_diagnostics": {
+            "type": "boolean",
+            "default": false,
+            "description": "Show Coq's \"Notice\" messages as diagnostics, such as `About` and `Search` output."
+          }
+        }
+      },
+      {
+        "title": "Goals and Info Panel",
+        "type": "object",
+        "properties": {
+          "coq-lsp.goal_after_tactic": {
+            "type": "boolean",
+            "default": false,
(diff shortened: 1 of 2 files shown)


editor/code/package.json
```diff
@@ -173,5 +190,11 @@
     "typecheck": "tsc -b",
     "compile": "npm run esbuild",
     "watch": "npm run esbuild -- --watch"
+  },
+  "__metadata": {
```
ejgallego, line 194: What is this?

editor/code/package.json
```diff
@@ -143,11 +136,35 @@
               "Show on click and on cursor movement",
               "Show on click, cursor, and command movement"
             ]
+          }
+        }
+      },      {
+        "title": "Language Server",
```
ejgallego, line 142: Maybe "Language Server Binary" ?

editor/code/package.json
```diff
@@ -126,6 +108,17 @@
             "type": "boolean",
             "default": false,
             "description": "Show Coq's \"Notice\" messages as diagnostics, such as `About` and `Search` output."
+          }
+        }
+      },
+      {
+        "title": "Goals",
```
ejgallego, line 115: Maybe "Info and Goals Panel"?

editor/code/package.json
```diff
@@ -75,33 +75,20 @@
     ],
     "configuration": [
       {
+        "title": "General",
```
ejgallego, line 78: Maybe "Proof Checking" or "Error Recovery" or maybe "Proof Checking and Error Recovery"?

ejgallego (commented):
Nice! Some suggestions in comments.

editor/code/package.json
Alizter: VSCode generated it by accident, I'll remove it

Alizter:
image

Alizter:
@ejgallego I added a changelog.

ejgallego:
How do I get to test this? I can't see where the hierarchy appears.

Alizter:
@ejgallego In the settings panel under extensions. You might need to reload vscode at least twice.

ejgallego (approved):
Thanks a lot this is great!

editor/code/package.json
```diff
@@ -173,5 +191,11 @@
     "typecheck": "tsc -b",
     "compile": "npm run esbuild",
     "watch": "npm run esbuild -- --watch"
+  },
+  "__metadata": {
```
ejgallego, line 195: Umm, still this keeps appearing here, but it should not, right?

ejgallego:
> @ejgallego In the settings window under extensions. You might need to reload vscode at least twice.

Ok got it, turns out you need to make the setting window large enough.

editor/code/package.json
Alizter: It's from `make ts-fmt`. It is quite annoying that compiling the project using `make extension` messes up all the formatting meaning I have to do `make ts-fmt` afterwards too, and that adds random crap here.

ejgallego: I don't see any of that; but again you should never do `make extension` or `make ts-fmt` if you have vscode open.

I'd bet what you see here is a race between some watch target in vscode with some extensions you are using.

ejgallego: BTW there is not way that make extension could modify any file in the source tree.

Issue #186, opened 2023-01-17
[infoview] Refine infoview protocol handling


ejgallego:
- check version
- validate input/output messages
- request status feedback

PR #215, opened 2023-01-22
[info panel] Refactor panel to proper object.

ejgallego:
We now handle the lifetime properly, and decouple the client from the lifetime of the panel.

This is a contribution towards #186

### DIFF
--- a/editor/code/src/client.ts
+++ b/editor/code/src/client.ts
@@ -4,13 +4,10 @@ import {
   extensions,
   ExtensionContext,
   workspace,
-  ViewColumn,
-  Uri,
   TextEditor,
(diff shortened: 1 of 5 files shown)


editor/code/src/client.ts
```diff
@@ -45,29 +42,11 @@ interface CoqLspClientConfig {
 
 let config: CoqLspClientConfig;
 let client: LanguageClient;
-let goalPanel: GoalPanel | null;
+// Lifetime of the goal panel == extension lifetime.
```
Alizter, line 45:
```suggestion
// Lifetime of the info panel == extension lifetime.
```

PR #217, opened 2023-01-22
[client] debug setting

Alizter:


CHANGES.md
```diff
@@ -12,6 +12,8 @@
    removed, beware of your setup in the extension settings
    (@ejgallego, #208)
  - Settings for the VSCode extension are now categorized (@Alizter, #212)
+ - Debug flags are now configurable in the VSCode extension
+   (@Alizter, #214)
```
Alizter, line 16:
```suggestion
   (@Alizter, #217)
```

ejgallego (commented):
I am not sure why we should bother the user with these flags, which are really internal.

Is there a way to hide these options by default?

Already the clutter in the settings panel is not good, so I'm unsure about this, seems like something it will impact all users a lot while the developers will barely benefit.

ejgallego:
I guess we can expose a single `debug` flags, and this flag would for now only turn backtraces on, I don't see how the rest could be useful. WDYT?

Alizter:
We don't have to have these settings available in the UI at all. We could just allow users to type the settings directly in their config. There is precedent for this kind of behavior in other extensions.

ejgallego:
> We don't have to have these settings available in the UI at all. We could just allow users to type the settings directly in their config. There is precedent for this kind of behavior in other extensions.

If that can be done then it's good. If not we should IMO only expose a generic `debug`.

Alizter:
@ejgallego This now works, you can configure the debug settings in your settings.json. The UI only shows a small debug entry which tells you to edit from settings.json.

I had to change the order in which coq was initialized so it came after recieving the init settings from the config.

controller/coq_lsp.ml
```diff
@@ -544,9 +541,17 @@ let lsp_main bt coqlib vo_load_path ml_include_path =
...
+    (* Core Coq initialization *)
+    let cmdline =
+      { Coq.Workspace.CmdLine.coqlib; vo_load_path; ml_include_path }
+    in
+
     (* LSP Server server initialization *)
     let workspace = lsp_init_loop ic oc ~cmdline in
 
+    let fb_queue = Coq.Protect.fb_queue in
+    let root_state = coq_init ~fb_queue ~bt in
```
ejgallego, line 553: The order here has to go back to `coq_init` first, then workspace init after.

That's not a big deal workspace initialization can take care of the debug stuff.

But indeed backtraces are special, for example you often want them to be enabled before the lsp_init loop takes place.

Alizter: If I don't do this then backtraces are not picked up from the settings. What is your suggestion for this?

ejgallego: Have `Workspace.apply` set it too, you'll need to add the option to the particular workspace setting.

Alizter:
I've gone for a more direct way of setting backtraces. I put it in the loop so if the setting is changed for example after loading the config, it is able to trace.

Alizter:
Maybe I am not so happy with this in the end.

ejgallego:
> Didn't really understand your Workspace.apply suggestion.

Sorry I didn't explain it properly, see #247

ejgallego:
I'd suggest we resurrect this PR, but only exposing a backtraces / debug flag in the vscode side.

Alizter:
@ejgallego Feel free to take over.

Issue #189, opened 2023-01-18
Unicode input


Alizter:
It would be good to have a mechanism like CoqIDE has where you can press `\` and then type in a unicode character name. It is possible to use extensions to insert unicode characters, but they don't appear in the completion interface.

ejgallego:
> It is possible to use extensions to insert unicode characters, but they don't appear in the completion interface.

That should be fixed by these extensions, shouldn't it?

I'm unsure this is something that coq-lsp should mess with, it seems to me that unicode input is something global for a particular editor.

Alizter:
Its not coq-lsp who should do something about this, but rather the vscode extension.

ejgallego:
I don't see why the coq-lsp vscode extension should mess with this.

Alizter:
I think its something that will make the user experience nicer, but not something we should focus on.

Alizter:
I think providing an optional unicode input method in the VSCode extension is a worthwhile feature. We should at the very least make note of recommended extensions / fixes for completion. Try writing Greek letters in a Coq file you are editing in VSCode with coq-lsp. I have not yet been able to come up with a good workflow, other than copy and pasting.

ejgallego:
I have little experience with VS Code, and indeed good unicode input is essential these days; I'm not sure how it is supposed to work with VS Code, indeed a pointer to the current best practices would be great.

I still don't see why `coq-lsp` has to be in charge of that, even in jsCoq I put it under its separate extension.

Why is unicode input specific to coq-lsp again?

By the way, would we decide to implement this, that would be a server feature as it is the completion call that would need to be extended with the unicode symbols.

Doing it client-wise would be silly as it would amount to a fully independent extension on their own.

I would also not be surprised if the current unicode extensions for VS Code are bad and we'd like to write our own, but for sure that's better done as a separate extension that contributes a completion provider to all text buffers.

Alizter:
>Why is unicode input specific to coq-lsp again?

Every editor (for every proof assistant I know) provides a way of inputting unicode values specific to them.

>By the way, would we decide to implement this, that would be a server feature as it is the completion call that would need to be extended with the unicode symbols.

I'm not at all suggesting we modify the LSP. This is really a quality-of-life feature of the extension itself.

>Doing it client-wise would be silly as it would amount to a fully independent extension on their own.

Yes that is perhaps silly, only if there exists an extension today that can do it. If that is not the case, we should develop and recommend our own.

[...]

PR #223, opened 2023-01-23
[infoview] Port to React

ejgallego:
This is prerequisite to implement / fix effectively #143 and #216 for example; the current Info View rendering using strings doesn't scale anymore to the kind of DOM manipulation we need.

Why React? That was a careful choice for many reasons, in particular seems like the best fit for our use case (so far); moreover, 3rd party package support is the best.


Alizter:
Do we want this for 0.1.4? I think we should wait a little. Let's get #189 sorted and then ship 0.1.4. This is a big enough change to warrant its own version.

ejgallego:
> Do we want this for 0.1.4? I think we should wait a little. Let's get #189 sorted and then ship 0.1.4. This is a big enough change to warrant its own version.

Actually except for a some minor style fixes, this is ready, and important for #143 and other works, so I think this will ship in 0.1.4.

I agree that once #189 is ready we release 0.1.4; however I'm lost with the problems you are having there, what your editor is doing seems very strange.

So far the only explanation for your problems is that VS Code is sending a completion request for a document change before the change itself. I bet this is a clear LSP protocol violation if that's what's happening!

Have you tried with a pristine VS Code configuration? (no extensions) A few of the problems you mention could be due to some badly behaved extension you have installed.

Alizter:
@ejgallego Since @artagnon reported that it is working as intended, I concede that it is most likely an issue on my end. We just need to update the table there and we are ready. I still think that the react stuff should wait till the next release so we can test it out a bit more.

ejgallego:
> @ejgallego Since @artagnon reported that it is working as intended, I concede that it is most likely an issue on my end. We just need to update the table there and we are ready. I still think that the react stuff should wait till the next release so we can test it out a bit more.

Actually I'd rather have the simple React UI in the wild ASAP, before we do some more serious work on it which could make early issues much harder to diagnose. This PR has little potential for problems. Do you see any potential issue?

The unicode stuff is way more serious, the server is crashing, so I'd like to have it working on your end first; otherwise it has a large potential to create lots of havoc on users as it is something we'll ship enabled by default.

Moreover, that PR needs quite a bit of work (including an option to disable that), while this one is ready except for a minor CSS adjustement.

PR #229, opened 2023-01-24
[doc manager] Remove [yield] from [Doc_manager.check]

ejgallego:
This will help to make this module independent of the particular event system chosen, that will help with the jsCoq port too.

### DIFF
--- a/controller/doc_manager.mli
+++ b/controller/doc_manager.mli
@@ -16,10 +16,10 @@
 (************************************************************************)
 
 module Check : sig
-  (** Check a document, or yield if there is none pending; it will send progress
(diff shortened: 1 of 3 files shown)


controller/doc_manager.mli
```diff
@@ -19,7 +19,8 @@ module Check : sig
   (** Check a document, or yield if there is none pending; it will send progress
```
Alizter, line 19:
```suggestion
  (** Check a document, or [yield] if there is none pending; it will send progress
```

Alizter (approved):

Alizter (approved):

Issue #15, opened 2022-07-03
[testing] Server testing


ejgallego:
See https://code.visualstudio.com/api/language-extensions/language-server-extension-guide#testing-the-language-server for more information, we should add server tests.

PR #241, opened 2023-01-25
Fix typo in -impredicative-set parsing

artagnon:


controller/coq_lsp.ml
```diff
@@ -318,7 +318,6 @@ let dispatch_request ~method_ ~params : RAction.t =
   (* Lifecyle *)
   | "initialize" ->
     LIO.trace "dispatch_request" "duplicate initialize request! Rejecting";
-    (* XXX what's the error code here *)
```
ejgallego, line 321: Actually that's a real TODO, `-32600` is likely not the right code.

Why do we mix unrelated changes in PRs?

ejgallego (changes requested):
Thanks a lot for the fix, however the unrelated change needs to be removed (or an issue opened) as it is a real TODO.

ejgallego:
I guess a changelog would be also useful, users may wonder why it was not working before.

ejgallego:
Also we could add an entry for this in `examples` (or tests very soon I hope!)

artagnon:
We can't add an example, as it requires a `_CoqProject` to be at the root of the repository.

ejgallego (approved):

ejgallego:
> We can't add an example, as it requires a `_CoqProject` to be at the root of the repository.

I guess we can add subdirs in `example`, let's discuss about this in person as it is really related to #15

PR #250, opened 2023-01-28
[Makefile] coq configure prefix and add nostdlib make

Alizter:


Makefile
```diff
@@ -34,10 +41,10 @@ watch: coq_boot
 build-all: coq_boot
 	dune build $(DUNEOPT) @all
 
+
+COQ_CONFIGURE_PREFIX=$(PWD)/_build/install/default/
 coq/config/coq_config.ml:
```
ejgallego, line 46: Are you sure this is going to work? If we don't call configure anymore, the config file will be generate on the build phase, and I'm not sure that's gonna pick the right variable.

Umm, there is a typo here by the way (should be `vendor/coq`)

Alizter: @ejgallego I believe it is this variable that is picked. That is indeed a typo.

ejgallego: I'm afraid you broke the build:
```
coqlib is at: /home/egallego/research/coq-lsp/_build/default/vendor/coq/../install/default/lib/coq
```
which is incorrect; the variable can't be possibly picked.

Please go back to standard configure, make lazy variable semantics is a hell.

ejgallego: The error message Coq gives for bad coqlib is really terrible, it just does `exit 1` and prints to the console :S

Alizter: I really wish we didn't have to do any configure for Coq :(

ejgallego: Yes that's a bit PITA.

ejgallego: Fixed the typo directly in master, I had noticed the other day but forgot to fix it and instead put the fix in some branch not yet merged.

PR #251, opened 2023-01-28
[CI] don't build stdlib in CI

Alizter:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,7 +48,7 @@ jobs:
         run: opam exec -- make opam-deps
 
       - name: 🧱 Build coq-lsp
-        run: opam exec -- make build
+        run: opam exec -- make build-nostdlib
 
   build-nix:
     name: Nix


ejgallego:
That's not a big deal due to the dune cache.

My main question here is what will happen when we do testing, maybe in that case we are gonna need the stdlib around?

For example we should have a test that Pff can be checked.

Alizter:
@ejgallego We can then start building the stdlib as we see fit. For now it is pointless since we don't do anything with it.

ejgallego:
> @ejgallego We can then start building the stdlib as we see fit. For now it is pointless since we don't do anything with it.

Well as we are gonna use it in the test suite I don't see the point to remove it now.

Anyways the build of the stdlib is cached and only happens when we bump the submodule.

ejgallego:
I'd say let's not fix what ain't broken.

Alizter:
Nix MacOS is broken upstream, I reported it: https://github.com/NixOS/nix/issues/7702

Alizter:
OK

ejgallego:
Actually once we start testing Coq master directly we will face this problem; indeed this idea to refine the CI is good, but IMHO we want to add some tests to the CI first, then decide where we run the tests.

I see 2 scenarios:
- we test only the compilation for an OCaml version: in this case we can even just do `dune build @check`
- we run the server: in this case we need to build some .vo files.

But actually, maybe in the latest case we could just build the Prelude, and have `coq-lsp` dependency tracking build the stuff; however that will be costly for stuff like `Pff.v`.

In fact one of the reasons I like the submodule is the caching. Would we test `master` directly, indeed this PR would be a must as we'd be wasting a lot of build time.

I'm unsure how to best solve this problem, but I'm very happy with the part of #250 for the nostdlib make if you want to add it. Also if you wanna use in CI why not, but as I pointed out, the build is already cached so we will gain nothing.

Following coq master seems indeed like a headache.

ejgallego:
I suggest we discuss CI in next meeting, if you are interested of course.

PR #254, opened 2023-01-28
[lsp] Cleanup LSP stuff, improve static typing.

ejgallego:
We can do much better, but we will need to introduce the Lang library first. For now, we remove the ppx yojson and import from the controller, and consolidate some types.

Attention to the `extra` stuff in diagnostics, we may need to reintroduce the --std flag.

cc: #2


ejgallego:
Pushed this to reduce some noise in #143

PR #270, opened 2023-01-29
[doc] [completion] Improve UTF-8 and position validation

### LINKED ISSUE #264: `\` throws an exception due to stale document from wrong order of didChange and completion
As #219 has been merged, we should document that the trigger character doesn't work for everybody (me). The work around is to press Ctrl + Space. There seems to be an ordering problem with the didChange and completion notification which is causing the document in the completion logic to be stale and causing havoc there.

ejgallego:
We are now more strict, and require the [Utf8] clients to take an action when the conversion fails.

Usually clients doing the conversion for the range will want to select the total length (last character) whereas the clients trying to grab a character will want to abort whatever they are doing.

I have reviewed all such uses, and found a couple of errors.

This also fixes the bug with the `fileProgress` range due to the above.

Re Fixes #264

### DIFF
--- a/controller/rq_completion.ml
+++ b/controller/rq_completion.ml
@@ -62,23 +62,23 @@ let unicode_list point : Yojson.Safe.t list =
   (* Coq's CList.map is tail-recursive *)
   CList.map (mk_unicode_completion_item point) ulist
 
-let get_line_at_point ~(doc : Fleche.Doc.t) ~point =
-  (* Guard against bad line number *)
-  let line, _char = point in
-  let line = min line doc.contents.last.line in
-  let line = Array.get doc.contents.lines line in
-  if Fleche.Debug.completion then Lsp.Io.trace "completion" ("line: " ^ line);
-  line
+let validate_line ~(doc : Fleche.Doc.t) ~line =
+  if Array.length doc.contents.lines > line then
+    Some (Array.get doc.contents.lines line)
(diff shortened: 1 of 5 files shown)


Alizter:
No exception anymore, but we are silently failing to serve the completion request because of the issue in #264, but I guess it doesn't cause any issues when pressing Ctrl + Enter. Interestingly, if you type `\` quickly during LSP initialization, then the request does get served correctly.

Alizter:
So I don't know if we should say it fixes #264, but rather makes it fail cleanly.

ejgallego:
Thanks for the testing Ali, I'm glad to hear this works better. Better typing will indeed help in the future.

> So I don't know if we should say it fixes #264, but rather makes it fail cleanly.

Yeah, I was thinking about the same. On the other hand, if the client does something inconsistent like in this case, what can we do on our side? I don't see any better way TBH.

I'd be happy to keep #264 open if there is some todo in the `coq-lsp` part, but if we cannot improve anything on our side, IMHO no point on keeping that bug open.

For now, I'd reopen the upstream bug but in the vscode repos.

ejgallego:
I'll then merge this for now, happy to have #264 reopened if we find a way to improve the server.

I'll add this problem to the known issues list in the README (at some point to be moved to a different place, but I'd like that list to be very visible for now)

PR #287, opened 2023-02-06
[fleche] Reject opening multiple Coq documents for buggy Coq versions.

### LINKED ISSUE #275: Reject checking multiple documents when Coq < 8.18
This is indeed very frustating for users, so indeed let's just do this for 8.18, and produce an error message for 8.17 and 8.16.

### LINKED ISSUE #281: Syntax error for "assert ... by ... "
assert ... by.. tactic generates "syntax error".
E.g. :

```coq
Lemma aaa : 1 = 0+1.
Proof.
assert (0+1 = 1) by (simpl; auto).
{ auto. }
Qed.
```

ejgallego:
Fixes #275 , fixes #281 

We use a heuristic in the fixed up branches as follows: if the Coq version contains LSP, then we assume the fix has been backported.

That should work when doing `opam pin`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,9 @@
    initialization parameter, moreover it ignored `rootUri` if present
    which goes against the LSP spec (@ejgallego, #295, report by Alex
    Sanchez-Stern)
+ - `coq-lsp` will now reject opening multiple files when the
+   underlying Coq version is buggy (@ejgallego, fixes #275, fixes
+   #281)
 
 # coq-lsp 0.1.4: View
 ---------------------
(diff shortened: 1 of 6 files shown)


ejgallego:
I will merge this (and release 0.1.5) when the branches noted in the README are ready.

ejgallego:
@Alizter what do you think of this?

Alizter (approved):

Alizter:
Logic seems sound, but I haven't tested it myself.

ejgallego:
> Logic seems sound, but I haven't tested it myself.

Thanks for the feedback @Alizter , indeed logic seems to work, I have done some testing; I hope the error message is Ok for users.

I'm waiting to see what's the outcome of https://github.com/coq/coq/pull/17200 just in case we would only need this for the 8.16 branch.

ejgallego:
Merging now, we can adjust the notes once we discuss about the Coq backport, before release (no point on holding this due to code conflicts)

PR #298, opened 2023-02-07
[doc] Add information about accepted URI types.

ejgallego:
Thanks to Alex Sanchez-Stern for the report / comment.

### DIFF
--- a/etc/doc/PROTOCOL.md
+++ b/etc/doc/PROTOCOL.md
@@ -50,6 +50,22 @@ If a feature doesn't appear here it usually means it is not planned in the short
 |-----------------------------------|---------|------------------------------------------------------------|
 
 
+### URIs accepted by coq-lsp
+
+`coq-lsp` only accepts `file:///` URIs; moreover, the URIs sent to the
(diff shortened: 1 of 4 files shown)


etc/doc/PROTOCOL.md
```diff
@@ -93,12 +109,39 @@ interface GoalAnswer<Pp> {
...
-
 ```
 
 which can be then rendered by the client in way that is desired.
 
-`proof/goals` was first used in the lambdapi-lsp server implementation, and we adapted it to `coq-lsp`.
+The main objects of interest are:
+- `Hyp`: This represents a pair of hypothesis names and type,
+  additionally with a body as obtained with `set` or `pose` tactics
+- `Goal`: Self-explanatory
```
Alizter, line 119: Is it the current goal or the list of goals?

Alizter (approved):

etc/doc/PROTOCOL.md
ejgallego: Added a bit more info, thanks for the review.

PR #301, opened 2023-02-07
[server] [lsp] Fix bug when parsing client option for unicode completion

ejgallego:
Yojson ppx doesn't work too well for this kind of enumeration types (they are serialized to a list with a string)

This was broken by #254


Alizter:
Can you report it to uojson upstream?

ejgallego:
> Can you report it to uojson upstream?

This is actually not a bug, it is intended to be that way, I understand.

Issue #187, opened 2023-01-17
[server] URI validation


ejgallego:
This is helpful in case clients do a mistake, as otherwise the error message will result confusing.

PR #302, opened 2023-02-07
workspace: support unicode filenames

artagnon:
Remove ad-hoc handling of uri, and use the Uri module instead. This gives us the benefit of being able to percent-decode, and support unicode filenames.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,7 @@
    #281)
  - Fix bug when parsing client option for unicode completion
    (@ejgallego #301)
+ - Support unicode characters in filenames (@artagnon, #302)
 
 # coq-lsp 0.1.4: View
 ---------------------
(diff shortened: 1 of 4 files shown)


ejgallego:
Great, note this likely fixes #187 

What is the story here if the URI is erroneous? Does that library something that returns a `Result.t` ?

artagnon:
The library returns an empty Uri.t on error:

https://github.com/mirage/ocaml-uri/blob/master/lib/uri.ml#L1105-L1112

It doesn't yet fix #187, as we need error handling code if the parse result is empty.

ejgallego:
Indeed, let's introduce a proper OCaml URI type in 0.1.6 and solve #187, thanks a lot for the fix Ram!

By the way folks, you know of a better way to avoid these annoying CHANGES.md conflicts?

ejgallego (approved):
Looks good to me, tho we want to push this check to the controller layer.

Note that in rq_init we have another instance of bad parsing, maybe that is breaking directories with unicode.

Alizter (approved):

PR #307, opened 2023-02-08
[code] Improve markdown syntax highglighting

4ever2:
Make `.mv` files inherit markdown grammar and inject coq grammar into coq codeblocks in both `.mv` and `.md` files.

### DIFF
--- a/editor/code/CHANGELOG.md
+++ b/editor/code/CHANGELOG.md
@@ -1,3 +1,8 @@
+# coq-lsp 0.1.5: Form
+----------------------
+
+- Improved syntax highligting of coq markdown files
+
 # coq-lsp 0.1.4: View
 ----------------------
 
--- a/editor/code/package.json
+++ b/editor/code/package.json
@@ -46,6 +46,16 @@
           ".mv"
         ],
         "configuration": "./coq.configuration.json"
+      },
+      {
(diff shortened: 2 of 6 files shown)


ejgallego:
That's just fantastic @4ever2 , thanks!

Would you mind adding an entry to the changes file ?

4ever2:
> Would you mind adding an entry to the CHANGES.md file ?

Added a changelog entry.

One limitation worth noting is that `.mv` files won't inherit grammar extension made by other VS Code extensions. As far as I can tell there is no good way around this limitation.

ejgallego (approved):

ejgallego:
> Added a changelog entry.

Thanks, actually the main changelog is at the root, but indeed the current procedure is confusing; I've updated `CONTRIBUTING.md` and `CHANGES.md`, please let me know if the entry in `CHANGES.md` looks Ok for you.

> One limitation worth noting is that .mv files won't inherit grammar extension made by other VS Code extensions. As far as I can tell there is no good way around this limitation.

Do you think this warrants opening an issue in VS Code upstream?

ejgallego:
Yes, that's a very interesting discussion point, and moreover here we have plans to allow the client to treat Coq-specific notations as sub-languages.

Issue #305, opened 2023-02-08
Support find in the goal window


Alizter:
We should support find (Ctrl + f) in the goal window. When you have a large goal, you sometimes might want to search through it.

Current workaround is copy paste the error message in a new buffer.

ejgallego:
Yes, actually a first step would be to make lists filtrable like in Lean; search like that may be a bit challenging as we don't have any code context, and the UI toolkit (https://github.com/microsoft/vscode-webview-ui-toolkit) is kinda bare.

Alizter:
Doesn't VS Code already have a search function? Do you know if it is possible to use?

ejgallego:
The info view panel is a barebones webview, so all this have to be built from scratch.

Alizter:
@ejgallego Not true, see my PR: https://github.com/ejgallego/coq-lsp/pull/309

ejgallego:
> @ejgallego Not true, see my PR: #309

Oh, indeed, well spotted!

ejgallego:
I think that other stuff such as the cursor is not possible without view-side code tho.

PR #309, opened 2023-02-09
Enable find widget in goal panel

Alizter:
Fix https://github.com/ejgallego/coq-lsp/issues/305

Tested and works.

### DIFF
--- a/editor/code/src/goals.ts
+++ b/editor/code/src/goals.ts
@@ -31,7 +31,7 @@ export class InfoPanel {
   }
 
   panelFactory() {
-    let webviewOpts: WebviewOptions = { enableScripts: true };
+    let webviewOpts = { enableScripts: true, enableFindWidget: true };
     this.panel = window.createWebviewPanel(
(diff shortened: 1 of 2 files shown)


editor/code/src/goals.ts
```diff
@@ -31,12 +31,14 @@ export class InfoPanel {
   }
 
   panelFactory() {
-    let webviewOpts: WebviewOptions = { enableScripts: true };
```
Alizter, line 34: This was an intersection type so you couldn't add enableFindWidget here directly. I've inlined it but the type is really `WebviewOptions & WebviewPanelOptions`.

ejgallego: You can also just remove the annotation (or add the intersection to the type annotation)

ejgallego (approved):
Oh, well spotted! Thanks a lot.

ejgallego:
There is a prettier failure, I will rebase, uninline, and add a changes entry.

PR #319, opened 2023-02-14
[doc] [error] Auto-admit bullets on new bullet error.

### LINKED ISSUE #300: Auto admit goal selectors
It would be nice if we could have more fine grained auto admits.

For example the following proof is incomplete:

```coq
Goal nat * nat * nat.
Proof.
  refine (_,_,_).
  - apply S.
  - apply S.
    exact 1.
  - exact 1.
Qed.
```
A use might want to work on the middle branch but the first branch prevents this from happening. coq-lsp should detect these situations and automatically insert an `admit.` so that the goal selector is properly closed. Something similar already happens in Coq by default when you do `{ incompelte. }`, using the bracket selectors.

ejgallego:
This is very useful when doing bulleted proofs.

Fixes #300

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -57,6 +57,10 @@
    #174, fixes #121, #122)
  - [internal] Error recovery can now execute full Coq commands as to
    amend states, required for #319 (@ejallego, #320)
+ - Auto-admit the previous bullet goal when a new bullet cannot be
+   opened due to an unsolved previous bullet. This also works for {}
+   focusing operators. This is very useful when navigating bulleted
+   proofs (@ejgallego, @Alizter, #319, fixes #300)
 
 # coq-lsp 0.1.4: View
 ---------------------
--- a/examples/goals.v
+++ b/examples/goals.v
@@ -14,3 +14,28 @@ Proof.
 pose (a := 3).
 exact a.
 Qed.
+
+Lemma failingBullet : (1 = 1) /\ (21 = 21 /\ 22 = 22) /\ (3 = 3).
+Proof.
+split;[|split].
+- 
+-
+-
+Qed.
+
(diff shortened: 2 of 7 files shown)


CHANGES.md
```diff
@@ -55,6 +55,9 @@
    will now be properly represented, as well as constructors for
    inductive types, projections for records, etc...  (@ejgallego,
    #174, fixes #121, #122)
+ - Auto-admit bullet goal when a new bullet cannot be opened due to
```
Alizter, line 58:
```suggestion
 - Auto-admit bullet goals when a new bullet cannot be opened due to
```

CHANGES.md
```diff
@@ -55,6 +55,9 @@
...
+   unsolved previous bullet, this is very useful when navigating
```
Alizter, line 59:
```suggestion
   an unsolved previous bullet, this is very useful when navigating
```

Alizter:
Can you add an example of a bulleted proof that demonstrates this behaviour?

CHANGES.md
ejgallego: Thanks.

CHANGES.md
ejgallego: Thanks.

ejgallego:
> Can you add an example of a bulleted proof that demonstrates this behaviour?

Last proof in goals.v should showcase the behavior.

examples/goals.v
```diff
@@ -13,4 +13,12 @@ Lemma bar : nat.
...
 exact a.
+Qed.
+
+Lemma failingBullet : (1 = 1) /\ (21 = 21 /\ 22 = 22) /\ (3 = 3).
+Proof.
+split;[|split].
+- 
+- 
+- auto.
 Qed.
```
Alizter, line 24:
```suggestion
Qed.

```

Alizter (approved):

PR #349, opened 2023-02-15
[lang] Move Ast Info to lang.

ejgallego:
This will make code more uniform.

### DIFF
--- a/controller/rq_completion.ml
+++ b/controller/rq_completion.ml
@@ -28,10 +28,10 @@ let grab_definitions ~doc ~f =
   List.filter_map df asts |> List.concat |> List.filter_map f
 
 let build_doc_idents ~doc : Yojson.Safe.t list =
-  let f { Coq.Ast.Info.name; _ } =
+  let f { Lang.Ast.Info.name; _ } =
     match name.v with
-    | Anonymous -> None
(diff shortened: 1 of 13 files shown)


controller/rq_completion.ml
```diff
@@ -28,10 +28,10 @@ let grab_definitions ~doc ~f =
   List.filter_map df asts |> List.concat |> List.filter_map f
 
 let build_doc_idents ~doc : Yojson.Safe.t list =
-  let f { Coq.Ast.Info.name; _ } =
+  let f { Lang.Ast.Info.name; _ } =
     match name.v with
-    | Anonymous -> None
-    | Name id -> Some (mk_completion ~label:Names.Id.(to_string id) ())
+    | None -> None
+    | Some id -> Some (mk_completion ~label:id ())
```
Alizter, line 34: why not do `Some label`?

Alizter (approved):

controller/rq_completion.ml
Alizter: Not a big deal for the moment.

ejgallego: Actually there is not a 1-1 mapping from ids to labels in fancier schemes, so maybe it is good like this, but indeed, not a big deal.

PR #351, opened 2023-02-15
[protocol] New request `coq/getDocument` to get serialized document.

ejgallego:
cc: https://github.com/ejgallego/coq-serapi/pull/318

### DIFF
--- a/editor/code/package.json
+++ b/editor/code/package.json
@@ -88,6 +88,10 @@
       {
         "command": "coq-lsp.goals",
         "title": "Coq LSP: Show Goals at Point"
+      },
+      {
+        "command": "coq-lsp.document",
+        "title": "Coq LSP: Get document in serialized format, print in console"
       }
     ],
     "keybindings": [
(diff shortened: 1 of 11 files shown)


ejgallego:
Tested, seems to work well; let's wait for some more feedback from Alectryon devs.

editor/code/package.json
```diff
@@ -88,6 +88,10 @@
       {
         "command": "coq-lsp.goals",
         "title": "Coq LSP: Show Goals at Point"
+      },
+      {
+        "command": "coq-lsp.document",
+        "title": "Coq LSP: Get document in serialized format, print in console"
```
Alizter, line 94: Do we want to expose this in the client?

ejgallego:
@cpitclaudel this seems to work fine in my testing, modulo a JSON serialization todo in `serlib` upstream, what do you think?

editor/code/package.json
ejgallego: Not sure, maybe we could indeed safe the stuff in a file. For now I need it for testing easily.

cpitclaudel:
Looks good! Thanks a lot :)

ejgallego:
Ok thanks, will add to `main` for now, I guess more API will have to be exposed, but for now this should allow you to start, and also includes completion status.

The current model of the server is that it will only run the request when the full document has checked, but a position could be included, which would mean "check to this point, run `getDocument`"

PR #360, opened 2023-02-16
[misc] Use dune build info for refined version view.

ejgallego:
For example on pins.

### DIFF
--- a/controller/lsp_core.ml
+++ b/controller/lsp_core.ml
@@ -274,7 +274,14 @@ let log_workspace w =
   LIO.trace "workspace" "initialized" ~extra;
   LIO.logMessage ~lvl:3 ~message
 
-let version () = Version.server
+let version () =
+  let dev_version =
+    match Build_info.V1.version () with
+    | None -> "NA"
+    | Some bi -> Build_info.V1.Version.to_string bi
+  in
+  Format.asprintf "version %s, dev: %s, Coq version: %s" Version.server
+    dev_version Coq_config.version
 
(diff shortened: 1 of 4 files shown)


Alizter:
@ejgallego Nix now works.

Alizter (approved):

controller/lsp_core.ml
```diff
@@ -274,7 +274,14 @@ let log_workspace w =
   LIO.trace "workspace" "initialized" ~extra;
   LIO.logMessage ~lvl:3 ~message
 
-let version () = Version.server
+let version () =
+  let dev_version =
+    match Build_info.V1.version () with
+    | None -> "NA"
```
Alizter, line 280:
```suggestion
    | None -> "N/A"
```
This is a de facto standard meaning "not applicable" or "not available".

flake.nix
```diff
@@ -49,7 +49,7 @@
           in
             l.attrValues {
               inherit serapi;
-              inherit (ocamlPackages) yojson cmdliner uri;
+              inherit (ocamlPackages) yojson cmdliner uri dune-build-info;
```
Alizter, line 52: I should note that this is really `dune-build-info.2.6.0`. I am pretty sure that it hasn't changed in 3.6 and we should be safe under the `V1` API anyway. It doesn't appear to cause any ill-effects in the Nix builds.

ejgallego:
Thanks a lot @Alizter , tweak to the message applied!

PR #361, opened 2023-02-16
[lsp] Preliminary support for push diagnostics

ejgallego:
- added `progress` and `partialResult` support to the core layer of the protocol
- pull diagnostics: wip

Note that VS Code won't set the partial progress token for `textDocument/diagnostic`, so this is not very useful for VS Code yet, but may be for other clients.


Alizter:
Did you mean to title it pull diagnostics?

ejgallego:
> Did you mean to title it pull diagnostics?

Yes, thanks!

PR #372, opened 2023-02-17
[Nix] build without submodules

Alizter:
This should allow greater composition with existing Coq nix setup.

Importantly we should build coq-lsp with the same version of OCaml and findlib that the Coq derivation is using.

WIP

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -19,28 +19,19 @@
       }: let
         l = lib // builtins;
 
-        coqVersion = "8.18";
+        coqVersion = "8.18+alpha";
         ocamlPackages = pkgs.ocamlPackages;
 
-        mkDunePackage = args @ {
-          pname,
-          src,
-          ...
-        }:
+        mkDunePackageFromInput = input: args:
           ocamlPackages.buildDunePackage (args
             // {
               duneVersion = "3";
-
-              inherit pname;
-              version = "${src.lastModifiedDate}+${coqVersion}";
-
-              src = src.outPath;
+              version = "${input.lastModifiedDate}+${coqVersion}";
+              src = input.outPath;
             });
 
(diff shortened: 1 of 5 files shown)


Alizter:
@r-muhairi  If you have some spare cycles, could you have a look at this? I think the issue is that we are not sharing enough with the Coq derivation (typically things like findlib). I want the build to use the latest Coq availalbe, perhaps as a flake input and share the OCaml version, findlib version etc. Then I believe the build will succeed.

Alizter:
FTR I did get the build to start by sharing the ocaml and findlib propagating dep, but the API that coq_8_17 is obviously too old for this repo to build so we need to get a fresher version.

Alizter:
Thanks @r-muhairi!

flake.nix
```diff
@@ -90,8 +102,13 @@
       flake = false;
     };
 
+    coq = {
+      url = "github:ejgallego/coq/8e0314415cc3cd528fc81bf11075d294179f880a";
```
Alizter, line 106:
```suggestion
      url = "github:ejgallego/coq";
```

flake.nix
```diff
@@ -90,8 +102,13 @@
...
+      flake = false;
+    };
+
     coq-serapi = {
-      url = "github:ejgallego/coq-serapi/v8.17";
+      url = "github:ejgallego/coq-serapi/9c8e91fdad0182675e0cad34bdeba54e63db86cd";
```
Alizter, line 111:
```suggestion
      url = "github:ejgallego/coq-serapi";
```

ejgallego:
@Alizter do you plan to continue work on this?

Alizter:
No.

PR #378, opened 2023-02-17
[code] Provide status bar button to stop / start the server.

### LINKED ISSUE #209: Add command to disable the language server
It would be good if we could deactivate the entire extension using a command. Sometimes you might be editing a .v file without wishing to start coq-lsp, in these cases deactivating would be useful.

ejgallego:
Thanks Ali for all the feedback on this.

Closes #209

### DIFF
--- a/editor/code/src/client.ts
+++ b/editor/code/src/client.ts
@@ -9,6 +9,7 @@ import {
   TextEditorSelectionChangeKind,
   StatusBarAlignment,
   StatusBarItem,
+  ThemeColor,
 } from "vscode";
 import {
   LanguageClient,
@@ -226,9 +227,15 @@ export function activate(context: ExtensionContext): void {
 
   const updateStatusBar = () => {
     if (client && client.isRunning()) {
-      lspStatusItem.text = "coq-lsp (running)";
+      lspStatusItem.text = "$(check) coq-lsp (running)";
+      lspStatusItem.backgroundColor = undefined;
+      lspStatusItem.tooltip = "coq-lsp is running. Click to disable.";
     } else {
-      lspStatusItem.text = "coq-lsp (stopped)";
+      lspStatusItem.text = "$(circle-slash) coq-lsp (stopped)";
+      lspStatusItem.backgroundColor = new ThemeColor(
+        "statusBarItem.warningBackground"
+      );
+      lspStatusItem.tooltip = "coq-lsp has been disabled. Click to enable.";
     }
   };
 
@@ -244,7 +251,6 @@ export function activate(context: ExtensionContext): void {
 
   start();
 }
-
 export function deactivate(): Thenable<void> | undefined {
   if (!client) {
     return undefined;
--- a/editor/code/src/client.ts
+++ b/editor/code/src/client.ts
@@ -7,6 +7,8 @@ import {
   TextEditor,
   TextEditorSelectionChangeEvent,
   TextEditorSelectionChangeKind,
+  StatusBarAlignment,
+  StatusBarItem,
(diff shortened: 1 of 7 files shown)


editor/code/src/client.ts
```diff
@@ -70,11 +74,12 @@ export function activate(context: ExtensionContext): void {
   }
   const stop = () => {
     if (client && client.isRunning()) {
-      client.stop();
+      client.dispose(2000).then(updateStatusBar); // EJGA: is this in milliseconds?
```
Alizter, line 77: It seems to be milliseconds but I only read about
```
      const timeout = setTimeout(() => panel.dispose(), 5000);
```
doesn't look like dispose takes an arg.

ejgallego: I see this in my types file:
```
dispose(timeout?: number): Promise<void>;
```
similarly for `stop`

ejgallego:
This needs a chicken tooltip, and indeed, time maybe to put a logo for the project.

editor/code/src/client.ts
ejgallego: `panel` doesn't take the argument, indeed. This is only for the node languageserver client.

ejgallego:
Also I'm thinking we can do 3 modes:
- on
- off
- lazy

Do we have an issue about lazy checking?

The easiest way to do is to check nothing, then have the goal request check up to that point.

That has terrible latency properties however.

ejgallego:
@Alizter IMHO this is ready for a first step merge, however I'm happy to further refine it until you are happy with the feature.

A good side-effect this PR has is that clicking the button will free all the `coq-lsp` memory. Not the right fix but still a workaaround.

Alizter:
So clicking around in a .v files will still open the goals panel when the server is "stopped".

ejgallego:
Noted, thanks.

ejgallego:
Seems to work better now, feel free to push the icons here or in a different PR.

ejgallego:
Another thing we should do is to document this in the README, but not sure where / how

ejgallego:
Couple of notes:
- icons don't display without the registration in package.json, that icon is a bit dated
- I think we should keep the "Running" / "stopped" text, due to accesibility (and in fact there are visually impaired Coq users)
- don't commit the settings.json file (we can add it to .gitignore
- the commit message needs amend as I had moved the icon to the left already (indeed that is the place it should be)

Alizter:
Sorry I didn't mean to commit those icons. I didn't get the Coq one to work in the end and opted for some default available ones. I think those slipped in in a rebase. Let me force push

ejgallego:
There is still a race in the panel code: if we stop the server but a `proof/goals` request is on the fly, the panel will reopen again when the request is received.

I'd suggest opening an issue as that will require deeper changes than what we should do here.

Alizter:
It also appears as though the progress scroll bar issue is still persisting.

ejgallego:
> It also appears as though the progress scroll bar issue is still persisting. Open a large .v file and then disable the server when it is somewhat through to see this.

I can't reproduce, let me try again.

ejgallego:
Ok found the problem, the update is throttled, so we need to postpone the cleanup too.

ejgallego:
Gonna rebase and fix that.

ejgallego:
Ok, some extra tweaks done. This is ready for more testing / merge if we don't find additional problems.

Alizter (approved):
Perfect!

ejgallego:
Very nice, the code can use some more cleanup, but we'll get there as more features appear.

PR #380, opened 2023-02-18
[fleche] Cache document creation

### LINKED ISSUE #363: [fleche] Cache Doc.create
This is needed in large projects.

ejgallego:
Some clients, such as Visual Studio Code, may open / close buffers
aggressively when working, this creates some extra load as we don't
cache document creation.

This PR does cache it. In general this has a very low footprint on
memory by itself, as the root state is shared across all buffers, but
of course, the documents themselves could be a big load, but that's
already the case so we don't make things worse.

Fixes #363

[Test suite passes ;)]

### DIFF
--- a/controller/cache.ml
+++ b/controller/cache.ml
@@ -22,7 +22,7 @@ let memo_cache_file = ".coq-lsp.cache"
 
 let memo_save_to_disk () =
   try
-    Fleche.Memo.save_to_disk ~file:memo_cache_file;
+    (* Fleche.Memo.save_to_disk ~file:memo_cache_file; *)
     LIO.trace "memo" "cache saved to disk"
   with exn ->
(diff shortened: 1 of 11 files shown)


controller/cache.ml
```diff
@@ -22,7 +22,7 @@ let memo_cache_file = ".coq-lsp.cache"
 
 let memo_save_to_disk () =
   try
-    Fleche.Memo.save_to_disk ~file:memo_cache_file;
+    (* Fleche.Memo.save_to_disk ~file:memo_cache_file; *)
```
Alizter, line 25: Did you mean to comment these out?

ejgallego: Yes, the stuff in Memo was really dead code so I've removed it, I'd like to keep these ones as stubs for now tho.

Issue #368, opened 2023-02-16
Span context computation


ejgallego:
This is required for:
- #50 
- #225

PR #391, opened 2023-02-19
[hover] Fix precedence of hypothesis vs globals.

ejgallego:


Alizter:
Not a fan of using exceptions for flow control here, but there is not much we can do since the problem comes from Coq.

ejgallego:
This needs indeed to be rewritten, for now it is just to get the feature working even if hacky, there are a lot of cases to handle, but once we implement #368 this will be handled in a principled way.

Feel free to start refactoring this code.

PR #395, opened 2023-02-19
[fleche] Add children info to the table-of-contents

ejgallego:


Alizter:
I'm confused, was this to do with the document symbols?

ejgallego:
The table of contents is used for jump to definition. So now jumping to children works.

ejgallego:
A better name we will use is some kind of `Index.t`

ejgallego:
So in this sense we have the map `definition -> range` and the list of definitions.

PR #409, opened 2023-02-21
ci: add jest-based testing

artagnon:
We only have a couple of simple lifecycle tests at the moment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,10 @@ build: coq_boot
 check: coq_boot
 	dune build $(DUNEOPT) @check
 
+.PHONY: test
+test: build
+	cd test && npm i && npm test
+
 .PHONY: fmt format
 fmt format:
 	dune fmt $(DUNEOPT)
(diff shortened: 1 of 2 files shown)


ejgallego:
So happy to see that this works on Windows!

ejgallego (approved):

Makefile
```diff
@@ -22,6 +22,10 @@ build: coq_boot
 check: coq_boot
 	dune build $(DUNEOPT) @check
 
+.PHONY: test
+test: build
```
ejgallego, line 26: At some point this could be moved to dune, which is usually better in terms of rebuilds.

We would ideally makes the deps a bit finer grained here, so tests run much faster. For example:
```
test/node_modules: test/package.json
     cd test && npm i

test: test/node_modules coq-lsp.install vendor/coq/coq-stdlib.install
      cd test && npm test
```
etc, you get the idea folks.

PR #411, opened 2023-02-21
[client]  Server status icon will now react properly to fatal server errors

### LINKED ISSUE #399: Status bar interacts badly when coq-lsp is not found
Enabling and disabling of the status bar does not interact well when the coq-lsp binary is not found. In this case we should be in the disabled state and when a user tries to reenable we give a message about coq-lsp binary not being found.

ejgallego:
Clicking on it will try to restart the server which is IMO a good thing.

Fixes #399

### DIFF
--- a/editor/code/src/client.ts
+++ b/editor/code/src/client.ts
@@ -130,6 +130,11 @@ export function activate(context: ExtensionContext): void {
             TextEditorSelectionChangeKind.Command
           );
         }
+      })
+      .catch((error) => {
+        let emsg = error.toString();
+        console.log(`Error in coq-lsp start: ${emsg}`);
+        setFailedStatuBar(emsg);
       });
   };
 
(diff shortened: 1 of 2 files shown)


editor/code/src/client.ts
```diff
@@ -239,6 +244,14 @@ export function activate(context: ExtensionContext): void {
     }
   };
 
+  const setFailedStatuBar = (emsg: string) => {
+    lspStatusItem.text = "$(circle-slash) coq-lsp (client failed to start)";
```
Alizter, line 248:
```suggestion
    lspStatusItem.text = "$(circle-slash) coq-lsp (error)";
```
The tooltip should be enough. The recommendation is not to take up to much space.

ejgallego: I see, but I find `error` too generic, what error? As this is the "exceptional" case, IMHO it is fine to be a bit longer, but I hear your concern.

Actually the message is wrong, it is the server that failed.

ejgallego: Made it a bit shorter.

PR #414, opened 2023-02-21
doc_manager: diagnostic with severity 0 is forbidden by lsp

artagnon:
Report 1 instead, for error.


ejgallego (approved):
Ouch! Indeed it is time to introduce an enum in `Lang.Diagnostics` and maybe a smarter constructor.

Alizter:
I think a variant is best.

PR #431, opened 2023-02-24
[io] Abstract IO functions

ejgallego:
Turns out the transport layer for the worker case is directly over JS objects, we thus generalize the IO types accordingly to prepare for the worker version of coq-lsp.

### DIFF
--- a/controller/doc_manager.mli
+++ b/controller/doc_manager.mli
@@ -18,13 +18,13 @@
 module Check : sig
   (** Check pending documents, return [None] if there is none pending, or
       [Some rqs] the list of requests ready to execute after the check. Sends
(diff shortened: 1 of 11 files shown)


controller/doc_manager.mli
```diff
@@ -19,12 +19,12 @@ module Check : sig
   (** Check pending documents, return [None] if there is none pending, or
       [Some rqs] the list of requests ready to execute after the check. Sends
       progress and diagnostics notifications to [ofmt]. *)
```
Alizter, line 21: Need to update comment. Also how am I supposed to read `ofn`? "output format...?"

ejgallego: Thanks, comment updated. `ofn` means "output function"

PR #452, opened 2023-03-06
ci: Replace nix fmt with flake check

### LINKED ISSUE #449: [ci] Format job doesn't fail anymore on incorrectly formatted sources.
See build log https://github.com/ejgallego/coq-lsp/actions/runs/4339317524/jobs/7576789585

This seems to originate in #284 , cc @Alizter @r-muhairi

r-muhairi:
Closes #449

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -102,8 +102,8 @@ jobs:
       - run: npm ci
       - run: npx --yes vsce ls
 
-  treefmt:
-    name: Format
+  nix-flake-check:
+    name: Nix Flake Check
     runs-on: ubuntu-latest
     steps:
       - name: 🔭 Checkout code
@@ -112,5 +112,5 @@ jobs:
           submodules: recursive
       - name: ❄️ Setup Nix
         uses: cachix/install-nix-action@v18
-      - name: 📐 Format with alejandra, ocamlformat, prettier
-        run: nix fmt
+      - name: 📐 Run flake check
+        run: nix flake check


Alizter:
So `nix fmt` formats the nix files and `nix check` actually runs ftmtree (which in turn runs all our formatters)?

r-muhairi:
> So `nix fmt` formats the nix files and `nix check` actually runs fmttree (which in turn runs all our formatters)?

nix fmt runs the `formatter` output generated by treefmt.config.??? not sure what was the option

nix flake check runs all the `checks` outputs found in `nix flake show` set through for example checks.something = drv;

All checks set are run through `nix flake check` I don't personally know if it is possible to run a specific target, but yeah.

The available checks are for treefmt set by using treefmt flake module.

Alizter:
I reread the code we have an I am not entirely sure how ocamlformat is being run. Where do we run that? Is it just using the makefile?

Alizter:
As a test, could you push a commit messing up the formatting in an ml file, a typescript file and a nix file to see how the CI complains?

r-muhairi:
> I reread the code we have an I am not entirely sure how ocamlformat is being run. Where do we run that? Is it just using the makefile?

If you look at the checks output in `nix flake show` you'll see a new check added on enabling treefmt
treefmt-nix comes with a predefined check using your treefmt.config setup, so it's just basically running treefmt behind the scenes with a TOML generated from our nix setup and a couple of extra flags for just checking not formatting and etc.

> As a test, could you push a commit messing up the formatting in an ml file, a typescript file and a nix file to see how the CI complains?

Done

Alizter:
Excellent thanks!

ejgallego:
Thanks a lot folks!

PR #454, opened 2023-03-07
[hover] Pass implicit information to printer.

### LINKED ISSUE #448: feature request: Hover should show implicit arguments
When hovering over a term, all the arguments are currently printed. We should make it so that implicit arguments are printed with `{ }` so that they can easily be ignored by the user. I believe not printing them at all is also an option, but may be confusing.

ejgallego:
This follows what Coq's `About` does upstream.

Note that still the printing is not very satisfactory in some cases, but in this case it seems Coq's `About` has the same problem.

So I'd suggest to close #448 and open an issue upstream about general implicit argument printing, WDYT @Alizter ?

Fixes #448.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,8 @@
    (@ejgallego, #181)
  - Fix cases when workspace / root URIs are `null`, as per LSP spec,
    (#453 , reported by orilahav, fixes #283)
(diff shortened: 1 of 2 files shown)


Alizter:
Yes, feel free to generalise the issue.

ejgallego:
Ok, I don't have the time this week to research the printing problems upstream, but added to a TODO list.

Issue #441, opened 2023-03-01
Error message browser becomes non-visible when there are many goals.


ejgallego:
Reported by Tomás Díaz

Alizter:
Does it not have a scroll bar?

ejgallego:
Yes but the user needs to scroll to see the error.

Alizter:
OK, then in this case it is not really bug but a quality of life improvement. As we discussed before, we should have a divider.

ejgallego:
This was fixed by #455 IMO

Issue #435, opened 2023-02-24
Goal panel offset misaligned


Alizter:
![image](https://user-images.githubusercontent.com/8614547/221246828-d03492c9-4fe3-44cc-b5ae-6579af4ba6d7.png)

As you can see, the goal panel seems to be misaligned and the top is clipping. Probably needs some CSS tweaking.

ejgallego:
That's actually not css but JavaScript, we call `scroolIntoView` , indeed, the scroll heuristics could use some tuning.

PR #455, opened 2023-03-07
[code] ui: Add divider between errors and goals (#441)

TDiazT:
Addresses #441. 
The error panel uses as much space as possible (might be missing some overflow style?) whenever there are errors. The goals panel has scrolling in case it gets smushed.

There is still an issue with `scrollIntoView` which makes the panel flicker in general and it might still be missing a top padding (#435). 

Lmk what you think (the gif below is clipped at the bottom but it's displaying ok in the editor).

Mar-07-2023 17-04-46

### DIFF
--- a/editor/code/views/info/Info.tsx
+++ b/editor/code/views/info/Info.tsx
@@ -1,7 +1,5 @@
-import React, { useEffect, useState } from "react";
+import { useEffect, useState, lazy, Suspense } from "react";
 import { GoalAnswer, GoalRequest, PpString } from "../../lib/types";
-
-// import "./media/info.css";
 
 // Main panel components
 import { FileInfo } from "./FileInfo";
@@ -9,6 +7,15 @@ import { Goals } from "./Goals";
 import { Messages } from "./Messages";
 import { ErrorBrowser } from "./ErrorBrowser";
 import { Program } from "./Program";
+
+import "./media/info.css";
+
+// Dynamic import because the project uses CommonJS and the module is an ECMAScript module
+// Top level await is supported with other `module` options in tsconfig.json
+const VSCodeDivider = lazy(async () => {
+  const { VSCodeDivider } = await import("@vscode/webview-ui-toolkit/react");
+  return { default: VSCodeDivider };
+});
 
 // First part, which should be split out is the protocol definition, second part is the UI.
 function doWaitingForInfo(info: GoalRequest) {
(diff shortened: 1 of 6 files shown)


editor/code/views/info/Info.tsx
```diff
@@ -1,15 +1,22 @@
...
 import { Program } from "./Program";
 
+import "./media/info.css";
+
+// Dynamic import because the project uses CommonJS and the module is an ECMAScript module
+// Top level await is supported with other `module` options in tsconfig.json
+const VSCodeDivider = lazy(async () => {
+  const { VSCodeDivider } = await import("@vscode/webview-ui-toolkit/react");
+  return { default: VSCodeDivider };
+});
```
TDiazT, line 18: Don't know if there's a better way but it kept complaining about CommonJS and ECMAScript modules and this is the only way I got it working. Otherwise there might be other options to change in `tsconfig.json` but I'm not very familiar with the consequences.

editor/code/views/info/Info.tsx
```diff
@@ -64,12 +71,23 @@ export function InfoPanel() {
...
-    </FileInfo>
+    <div className="info-panel-container">
+      <div className="info-panel">
+        <FileInfo textDocument={goals.textDocument} position={goals.position}>
+          <Goals goals={goals.goals} />
+          <Program program={goals.program} />
+          <Messages messages={goals.messages} />
+        </FileInfo>
+      </div>
+      {!goals.error ? null : (
```
TDiazT, line 82: Moved the check here because this component is the one layering the others. The `ErrorBrowser` shouldn't be the one deciding whether it's displayed or not imo, but the parent that sets the structure.

editor/code/views/info/media/index.css
```diff
@@ -4,3 +4,7 @@ body {
   font-family: var(--vscode-editor-font-family);
   /* font-family: 'Segoe UI', Tahoma, Geneva, Verdana, sans-serif; */
 }
+
+#root {
+  height: 100vh;
```
TDiazT, line 9: Had to go with this to get the `overflow` working, don't know if it's the best approach (using 100% wouldn't work).

editor/code/views/info/Info.tsx
ejgallego: I got it working in `views/perf` just using:
```typescript
import {
  VSCodeDataGrid,
  VSCodeDataGridRow,
  VSCodeDataGridCell,
} from "@vscode/webview-ui-toolkit/react";
```
Where did you get the complaints?

editor/code/views/info/Info.tsx
ejgallego: I did understood that in React it was idiomatic for components to handle these kind of cases returning `null` by themselves, but this looks OK to me too.

ejgallego (approved):
LGTM thanks! Sorry for the delay in reviewing this.

editor/code/views/info/media/index.css
ejgallego: I see, can we add a test case in examples that overflows both the error panel and the goal panel?

This way we can tweak that.

PR #467, opened 2023-03-17
Reduce thread delay in process queue for a speedup on short commands

HazardousPeach:
Instructions to compare performance across branches, as well as timing numbers from a particular machine, are available here: https://pastebin.com/TzBKPj3W. Depending on machine, this seems to lead to 2-4x speedup.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,8 @@
+# coq-lsp 0.1.8
+---------------
+ - Decrease latency of requests by ~4x by reducing threading timer
+   (fix suggested by @ejgallego, implemented by @hazardouspeach)
+
 # coq-lsp 0.1.7: Just-in-time
 -----------------------------
 
--- a/controller/coq_lsp.ml
+++ b/controller/coq_lsp.ml
@@ -44,7 +44,7 @@ let rec process_queue ~ofn ~state =
        I/O *)
     exit 0
   | Some (Yield state) ->
-    Thread.delay 0.1;
+    Thread.delay 0.01;
     process_queue ~ofn ~state
   | Some (Cont state) -> process_queue ~ofn ~state
 


controller/coq_lsp.ml
```diff
@@ -44,7 +44,7 @@ let rec process_queue ~ofn ~state =
        I/O *)
     exit 0
   | Some (Yield state) ->
-    Thread.delay 0.1;
+    Thread.delay 0.01;
```
Alizter, line 47: Why do we delay at all actually?

HazardousPeach: Errrm, that's actually a pretty good question, I'm not sure. @ejgallego maybe you can answer that?

CHANGES.md
```diff
@@ -1,3 +1,8 @@
+# coq-lsp 0.1.8
+---------------
+ - Decrease latency of requests by ~4x by reducing threading timer
+   (fix suggested by @ejgallego, implemented by @hazardouspeach)
+
```
ejgallego, line 5: This can go in 0.1.7

controller/coq_lsp.ml
ejgallego: Otherwise when coq-lsp is idle we would eat all of the CPU busy-waiting.

We could use thread-aware queues, so the thread could be put to sleep and woken up by that, **however** that may work badly on Windows and jsCoq so I prefer this very simple method until our testing infra is better on these platforms.

Happy if someone wants to do an experiment with https://v2.ocaml.org/releases/4.14/api/Condition.html

ejgallego (commented):
Thanks @HazardousPeach , looks good to me, however I will tweak your fix slightly as I think this parameter should be settable by a command line option, a 0.01 delay might be too short for UIs (or not), so making this a parameter will allow users some freedom to tweak this and report back.

controller/coq_lsp.ml
ejgallego: But yes, that means with our setup we often get a 0.1 latency, as I wrote on the other comment we better make this a parameter and find the right spot (UI programming .... XD)

ejgallego:
Replaced by a command line arg in #471 , thanks!

PR #469, opened 2023-03-26
Added required support for ViZX Extension

bhaktishh:
ViZX can be found [here](https://marketplace.visualstudio.com/items?itemName=inQWIRE.vizx). I just added code to execute a ViZX command when the extension is active; the goal is to enrich some textual goal display with a visualization; this command will simply open a webview with the visualization. ViZX has to be installed and active for the command to be executed; the goals object is the only information passed to ViZX.

### DIFF
--- a/editor/code/src/goals.ts
+++ b/editor/code/src/goals.ts
@@ -1,4 +1,4 @@
-import { Position, Uri, WebviewPanel, window, ViewColumn } from "vscode";
+import { Position, Uri, WebviewPanel, window, ViewColumn, extensions, commands } from "vscode";
 import {
   BaseLanguageClient,
   RequestType,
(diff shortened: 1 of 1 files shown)


editor/code/src/goals.ts
```diff
@@ -78,6 +78,10 @@ export class InfoPanel {
   // notify the info panel that we have fresh goals to render
   requestDisplay(goals: GoalAnswer<PpString>) {
     this.postMessage("renderGoals", goals);
+    let vizx = extensions.getExtension("inQWIRE.vizx");
+    if (vizx?.isActive) {
+      commands.executeCommand("vizx.lspRender", goals);
+    }
```
ejgallego, line 84: Looks good to me as a prototype, will merge so you can continue your testing.

I propose we indeed move towards a registration mechanism like done in most parts of VSCode, so eventually we can replace this code with your extension doing in activation:
```typescript
let coq-lsp = extensions.getExtension("coq-lsp").?API_v1
coq-lsp.registerGoalProcessorString((goals) => lspRender(goals))
```

We may need a bit different API once we need to properly manage the UI resources both extensions handle, but that should be a good start.

ejgallego (approved):

ejgallego:
Couple of comments for further PRs:
- we setup github it prefers the commits to be signed, this is a kind of authorship check
- we format typescript code using prettier, (that was the CI failure, not a big deal but maybe to keep in mind for future PRs)
- while this change seems mostly internal, you are welcome to add an entry in the CHANGES.md file, IMHO we should add something about ViZX to changes, and if you want to the README

PR #470, opened 2023-03-29
[protocol] Include goal output type in goals request

ejgallego:
This is a step to make life of clients easier, cc: @bhaktishh

### DIFF
--- a/controller/lsp_core.ml
+++ b/controller/lsp_core.ml
@@ -28,6 +28,8 @@ let int_field name dict = U.to_int (field name dict)
 let dict_field name dict = U.to_assoc (field name dict)
 let list_field name dict = U.to_list (field name dict)
 let string_field name dict = U.to_string (field name dict)
+let ofield name dict = List.(assoc_opt name dict)
+let ostring_field name dict = Option.map U.to_string (ofield name dict)
 
 module LIO = Lsp.Io
 module LSP = Lsp.Base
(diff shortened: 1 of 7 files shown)


controller/lsp_core.ml
```diff
@@ -265,7 +267,26 @@ let do_position_request ~postpone ~params ~handler =
     (Request.Data.PosRequest { uri; handler; point; version; postpone })
 
 let do_hover = do_position_request ~postpone:false ~handler:Rq_hover.hover
-let do_goals = do_position_request ~postpone:true ~handler:Rq_goals.goals
+
+(* We get the format from the params *)
+let get_pp_format_from_config () =
+  match !Fleche.Config.v.pp_type with
+  | 1 -> Rq_goals.Pp
+  | _ -> Rq_goals.Str
```
Alizter, line 275: Do we want to do this or fail on pp types we don't know. I thinking people might use this to see if pp is supported, but it would give a false positive which might cause headaches.

Alizter: At the very least we should output a warning in the log.

ejgallego: Yes, it is a pity using deriving_yojson on such unitary types doesn't work as we'd like, hopefully we can improve this part of the serialization.

editor/code/src/goals.ts
```diff
@@ -1,4 +1,12 @@
-import { Position, Uri, WebviewPanel, window, ViewColumn, extensions, commands } from "vscode";
+import {
```
Alizter, line 1: This appears to be unrelated

ejgallego: Yes it was a formatting problem from previous PR, my bad.

It is too tiny to split IMO, so if you don't mind I'll push it here.

PR #480, opened 2023-04-07
Lower latency

HazardousPeach:
This pull request does two things to increase throughput of coq-lsp commands. It adds a `--concise` flag which suppresses the printing of most lsp messages, leaving only those necessary to make requests and get the results (no diagnostic, progress, info printing, etc). And it also changes json output to print on one line instead of pretty printing.

For a basic task of adding the statement "Theorem x: nat." to a document and then getting the goals, in my measurements the former gives a 10.6x speedup, and the latter gives another 1.89x speedup, for a total of 20.2x speedup.

If there's a problem with the json change for users using coq-lsp in an editor, that can be put behind the `--concise` flag too, just wasn't sure. It should be fine, but it does add another trailing newline to each message to make sure line buffering doesn't prevent us from seeing the message.

### DIFF
--- a/controller/coq_lsp.ml
+++ b/controller/coq_lsp.ml
@@ -51,7 +51,10 @@ let rec process_queue ~delay ~concise ~ofn ~state : unit =
 let concise_cb =
   Fleche.Io.CallBack.
     { trace = (fun _hdr ?extra:_ _msg -> ())
-    ; send_diagnostics = (fun ~ofn:_ ~uri:_ ~version:_ _diags -> ())
+    ; send_diagnostics =
+        (fun ~ofn ~uri ~version diags ->
+          if List.length diags > 0 then
+          Lsp.JLang.mk_diagnostics ~uri ~version diags |> ofn)
     ; send_fileProgress = (fun ~ofn:_ ~uri:_ ~version:_ _progress -> ())
     }
 
--- a/controller/coq_lsp.ml
+++ b/controller/coq_lsp.ml
@@ -31,10 +31,10 @@ let exit_message () =
 
 let lsp_cleanup () = exit_message ()
 
-let rec process_queue ~delay ~ofn ~state : unit =
+let rec process_queue ~delay ~concise ~ofn ~state : unit =
   if Fleche.Debug.sched_wakeup then
     LIO.trace "<- dequeue" (Format.asprintf "%.2f" (Unix.gettimeofday ()));
-  match dispatch_or_resume_check ~ofn ~state with
+  match dispatch_or_resume_check ~ofn ~concise ~state with
   | None ->
     (* As of now, we exit the whole program here, we could try an experiment to
        invert the threads, so the I/O routine is a thread and process_queue is
@@ -45,8 +45,18 @@ let rec process_queue ~delay ~ofn ~state : unit =
     exit 0
   | Some (Yield state) ->
     Thread.delay delay;
-    process_queue ~delay ~ofn ~state
-  | Some (Cont state) -> process_queue ~delay ~ofn ~state
+    process_queue ~delay ~concise ~ofn ~state
+  | Some (Cont state) -> process_queue ~delay ~concise ~ofn ~state
+
+let concise_cb =
+  Fleche.Io.CallBack.
+  { trace =
+      (fun _hdr ?extra:_ _msg -> ())
+    ; send_diagnostics =
+        (fun ~ofn:_ ~uri:_ ~version:_ _diags -> ())
+    ; send_fileProgress =
+        (fun ~ofn:_ ~uri:_ ~version:_ _progress -> ())
+  }
 
 (* Main loop *)
 let lsp_cb =
@@ -77,7 +87,7 @@ let rec lsp_init_loop ~ifn ~ofn ~cmdline ~debug =
     | Init_effect.Loop -> lsp_init_loop ~ifn ~ofn ~cmdline ~debug
     | Init_effect.Success w -> w)
 
-let lsp_main bt coqcorelib coqlib ocamlpath vo_load_path ml_include_path delay =
+let lsp_main bt coqcorelib coqlib ocamlpath vo_load_path ml_include_path delay concise =
   (* Try to be sane w.r.t. \r\n in Windows *)
   Stdlib.set_binary_mode_in stdin true;
   Stdlib.set_binary_mode_out stdout true;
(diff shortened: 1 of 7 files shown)


HazardousPeach:
Pushed a new commit so that errors are still reported in --concise mode, even though most publishDiagnostics messages are supressed

controller/doc_manager.ml
```diff
@@ -172,16 +174,17 @@ end = struct
...
       let target = get_check_target handle.pt_requests in
       let doc = Fleche.Doc.check ~ofn ~target ~doc:handle.doc () in
       let requests = Handle.update_doc_info ~handle ~doc in
-      send_diags ~ofn ~doc;
-      (* Only if completed! *)
-      if completed ~doc then send_perf_data ~ofn ~doc;
+      if not concise then (
+        send_diags ~ofn ~doc;
+        (* Only if completed! *)
+        if completed ~doc then send_perf_data ~ofn ~doc);
```
ejgallego, line 187: Maybe we could introduce finer grained options (and set them properly) as done for other configuration items?

I'd do that and remove the `concise` parameter, as we already have `Fleche.Config` for other options.

lsp/io.ml
```diff
@@ -61,9 +61,9 @@ let log = ref (fun _ _ -> ())
 let send_json fmt obj =
   Mutex.lock mut;
   if Fleche.Debug.send then !log "send" obj;
-  let msg = F.asprintf "%a" J.(pretty_print ~std:true) obj in
+  let msg = J.to_string obj in
```
ejgallego, line 64: Nice catch!

Let's keep the pretty printed version around if possible for a while (disabled under a flag), it is rare, but sometimes we need to do wire-level debugging and in that case the pretty printing helps.

ejgallego (approved):
Very nice, LGTM thanks!

Feel free to merge as is if you want, or to address some of the comments if you feel like.

I'd would personally remove the `~concise` parameter to `Doc_manager` and instead use flags in `config.ml`.

If you put all the options in the flags, you could even remove the callback setting conditional and just tweak said options on the init message.

Doing that makes harder to use the command line as config is overwritten at LSP init time, just saying.

ejgallego:
> Pushed a new commit so that errors are still reported in --concise mode, even though most publishDiagnostics messages are supressed

Mmm, I think this actually inverts the logic, the callback for diags is used to send incremental diags (which you don't want), it is the end of check diags call the one that should be kept.

Actually you could just use pull diagnostics if you want.

ejgallego:
I just realized that actually yes, having a flag to disable diagnostics at all is fine, as clients may want to enable pull diags and disable the old style async ones.

controller/coq_lsp.ml
```diff
@@ -45,8 +45,18 @@ let rec process_queue ~delay ~ofn ~state : unit =
...
     Thread.delay delay;
-    process_queue ~delay ~ofn ~state
-  | Some (Cont state) -> process_queue ~delay ~ofn ~state
+    process_queue ~delay ~concise ~ofn ~state
+  | Some (Cont state) -> process_queue ~delay ~concise ~ofn ~state
+
+let concise_cb =
+  Fleche.Io.CallBack.
+    { trace = (fun _hdr ?extra:_ _msg -> ())
+    ; send_diagnostics =
```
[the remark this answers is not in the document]
HazardousPeach: You mean the `send_diagnostics` callback? That appears to be where error messages from invalid tactics come back, so I want those. In my testing, this version appears to return all error messages while supressing anything that isn't an error or warning.

controller/doc_manager.ml
HazardousPeach: Yeah that makes sense! Especially for warnings, it's not clear whether --concise should include them or not, having the user be able to specify a verbosity level would be good. I'll look into how Fleche.Config works.

lsp/io.ml
HazardousPeach: Yeah, that makes sense. Once I understand how to use Fleche.Config it shouldn't be hard to add an extra parameter for json pretty printing, I was just cautious when I thought it meant threading another parameter through a bunch of function calls.

HazardousPeach:
>

HazardousPeach:
> I'd would personally remove the `~concise` parameter to `Doc_manager` and instead use flags in `config.ml`.
> 
> If you put all the options in the flags, you could even remove the callback setting conditional and just tweak said options on the init message.
> 
> Doing that makes harder to use the command line as config is overwritten at LSP init time, just saying.

Hmm, I'm not sure I understand what you mean by this, but it sounds important. Are you saying that if I use `Fleche.Config` instead of the parameters I'm using, then I won't be able to use the command line, and will have to edit a configuration file instead? Relatedly, why did you implement the `-D` parameter as a function parameter instead of inside Fleche.Config?

HazardousPeach:
> > Pushed a new commit so that errors are still reported in --concise mode, even though most publishDiagnostics messages are supressed
> 
> Mmm, I think this actually inverts the logic, the callback for diags is used to send incremental diags (which you don't want), it is the end of check diags call the one that should be kept.
> 
> Actually you could just use pull diagnostics if you want.

I'm confused, what do you mean by inverts the logic? When I tested this change, it correctly sent messages for errors and warnings, which is what I expected, while suppressing all other messages.

HazardousPeach:
I added another commit, because it seems like different errors are reported in different places. Most errors were reported through the Fleche callback, but "Anomaly" type messages were being pulled from fleche through `Fleche.Doc.Node.diags`  and `send_diags`. This new commit makes sure those errors are still shown in `concise` mode.

HazardousPeach:
This last set of commits responds to a few pieces feedback from Emilio. 

First of all, instead of a --concise flag passed on the command line, we add a new configuration option "verbosity", which can be set during lsp initialization. A verbosity of 2 is the default, which means the same as omitting the --concise flag. Verbosity levels of 0 and 1 are currently identical, both meaning the same thing as the --concise flag. I left two verbosity levels there because eventually verbosity 0 might mean "never report warnings" while verbosity 1 is "don't report progress, cache, etc, but report warnings".

Second of all, instead of pretty-printed json being removed entirely, there is now a "pp_json" configuration option which can also be set during lsp initialization. It's default is "false", meaning the non-pretty-printed json that was previously the only option on this branch. However, when set to "true", it pretty-prints json like done outside of this branch, to support wire-level debugging.

fleche/config.ml
```diff
@@ -37,6 +37,8 @@ type t =
         (** Pretty-printing type in Info Panel Request, 0 = string; 1 = Pp.t; 2
             = Coq Layout Engine *)
   ; show_stats_on_hover : bool [@default false]
+  ; verbosity : int [@default 2]
```
ejgallego, line 40: @HazardousPeach do you recall what the values do mean, I'd like to document them.

ejgallego (commented):
Sorry for the delay reviewing this, the changes LGTM thanks!

I'd like to add an entries in CHANGES.md , let me know if you have a suggested wording, otherwise I will take care of it.

It would also be good to have a test in `tests/` so we don't break this mode with the upcoming changes.

Motivated by this PR, I have been thinking quite a bit about coq-lsp configuration during my leave.

And indeed I think we want to introduce a bit more general configuration architecture, as to start with several clients do pass different capabilities to the server, so in this sense we need to be quite flexible on what we send back to the client (and avoid doing such work)

PR #493, opened 2023-05-15
Adapt to coq/coq#17605 (summary doesn't use marshallable argument when freezing)

SkySkimmer:


ejgallego:
Submodules bumped, will merge when CI passes.

PR #494, opened 2023-06-07
Adapt to coq/coq#17696 (exposed Mltop.load_module)

SkySkimmer:


ppedrot:
This is backwards compatible, right? So it can go already in.

ppedrot:
@SkySkimmer it seems you must sign your commit to be able to merge it here.

PR #502, opened 2023-06-20
update nix flake

Alizter:
Doesn't fix the macos issue #508 does that.


ejgallego:
Don't forget to update the meta-data when merging a PR.

Issue #487, opened 2023-04-17
`simpl` tactic isn't interrupted properly


HazardousPeach:
**Describe the bug**

For large or complex contexts, the `simpl` command can hang for a long time (indefinitely?) and isn't interruptible from coq-lsp.

**To Reproduce**
Since this requires complex contexts to reproduce, we'll reproduce it using CompCert.
1. Clone CompCert: `git clone git@github.com:AbsInt/CompCert.git`
2. Start the coq-lsp binary.
3. Send the `initialize` message, with the rootUri, rootPath, and workspace uri pointing to the toplevel of your CompCert clone.
4. Send the `initalized` message.
5. Send a `openDoc` message with an empty document.
6. Send a `didChange` message with the following document:
```
Require Import Coqlib.\nRequire Import Integers.\nRequire Import Values Memory.\nRequire Import Cminor CminorSel.\nRequire Import SelectOp.\nSection CMCONSTR.\nVariable ge: genv.\nVariable sp: val.\nVariable e: env.\nVariable m: mem.\nDefini tion unary_constructor_sound (cstr: expr -> expr) (sem: val -> val) : Prop :=\n  forall le a x,\n  eval_expr ge sp e m le a x ->\n  exists v, eval_expr ge sp e m le (cstr a) v /\\ Val.lessdef (sem x) v.\nLemma eval_mulimm_base:\n  forall n, unary_constructor_sound (mulimm_base n) (fun x => Val.mul x (Vint n)).\ninducti on n.\neconstructor.\nunfold mulimm_base.\nunfold Int.one_bits.\nunfold map.\nsimpl.
```
7. Wait a bit, noticing that the doc fails to finish checking.
8. Send another `didChange` message with the following document:
```
[...]

Issue #484, opened 2023-04-13
Typeclass resolution can lock coq-lsp responsiveness when looping


Alizter:
It is a common occurrence to tweak a typeclass definition causing later lemmas to loop when searching. When this happens, it causes coq-lsp to become unresponsive.

ejgallego:
Can you provide a test case?

Alizter:
Here is a test case:

```
From Coq Require Import Prelude.

Axiom Loop : Type.
Existing Class Loop.

Axiom loop : Loop -> Loop.
#[export] Existing Instance loop.

Global Instance foo : Loop.
Proof. Admitted.

Goal Loop.
exact _.
```

Try commenting out `foo` and typeclass search will loop. When this happens, LSP becomes unresponsive.

ejgallego:
Thanks a lot Ali, I have opened a file with such cases and I hope to resolve them soon.

ejgallego:
@Alizter note this example doesn't loop anymore, what'd be the suggested fix?

ejgallego:
I did 
```
Global Instance foo (a : Loop) : Loop.
Proof. Admitted.
```
for example, and that seems to work, but I don't know how TC work.

Alizter:
I yougive `foo` a lower priority with `| 10` that should work.

Issue #139, opened 2023-01-02
Closing a buffer should terminate coq-lsp


Alizter:
Coq-lsp should terminate when the buffer is closed. Otherwise it stays open and has to be restarted manually.

ejgallego:
`coq-lsp` server lifetime follows the LSP specification, it is the client who controls the lifetime; do you have a more concrete example?

Alizter:
Open T2048.v and then close the buffer. Coq lsp still runs for a bit.

ejgallego:
Ok will try; actually the most likely problem is that `T2048.v` gets deep into the kernel to check the `Qed`, such deep conversion loops don't do `check_for_interrupt` often so there you have your problem: the client sends the shutdown message properly, however the queue is not processed.

So if my guess is correct, the problem is more general, and the issue could be retitled as "Coq doesn't call `check_for_interrupt` often enough, and coq-lsp becomes unresponsive", and that's a kind of upstream issue.

Recall that at first I was skeptical that `check_for_interrupt` would work well enough, but actually it has!

We've been discussing with OCaml devs how to do this better (and have a check on allocation), but the tradeoffs here are complex.

I think for now we just need to try to gather all the cases where coq-lsp gets stuck, maybe just an extra check in Coq could solve this without an invasive signal-handling patch.

Alizter:
>So if my guess is correct, the problem is more general, and the issue could be retitled as "Coq doesn't call check_for_interrupt often enough, and coq-lsp becomes unresponsive", and that's a kind of upstream issue.

Yes, that is a better title. Updated.

>Recall that at first I was skeptical that check_for_interrupt would work well enough, but actually it has!

This is good news. Let us try to make better use of it then.

>We've been discussing with OCaml devs how to do this better (and have a check on allocation), but the tradeoffs here are complex.

Do you mean: https://github.com/ocaml/ocaml/issues/11411?

>I think for now we just need to try to gather all the cases where coq-lsp gets stuck, maybe just an extra check in Coq could solve this without an invasive signal-handling patch.

Let's definitely try that first.

ejgallego:
I did some more research on this example, the problem seems at:
```
#3956 0x000056489b651ac7 in camlDeclare__to_constr_body_6449 () at vendor/coq/vernac/declare.ml:1623
#3957 0x000056489b651bf7 in camlDeclare__fun_10409 () at vendor/coq/vernac/declare.ml:1648
```
[...]

PR #509, opened 2023-06-21
[coq] Support for token-based interruption

ejgallego:
This very preliminary PR reifies the Coq API so we can pass an interruption / cancellation token around.

Two implementations of the API are provided:
- one using the "classical" `Control.interrupt` variable present in Coq
- one using `memprof-limits` by @gadmm

The code is very preliminary and will likely be rewritten to better accomodate the API, however the current version works, and testing is much appreciated! This is a delicate part of the system so testing is badly needed.

In the examples we had, `memprof-limits` works superb! In particular it works in all the examples where polling fails :D 

I think this should allow us to get rid of Coq polling eventually.

Some open issues:
- While token-based interruption works pretty well, there are a few corner cases where the user experience is not optimal. In particular, when a command that takes a lot of time to run (like a `Qed`) is interrupted (for example the user does a search), we need to restart it from scratch. That's a big PITA. We could solve this in OCaml 5 both with Domains or algebraic effects + continuations, however....
- ... unfortunately `memprof-limits` is still not ported to OCaml 5, thus https://github.com/ocaml/ocaml/issues/11411 is still open.

Also, it didn't take long for us to find a bug in Coq and how they handle resources, c.f. https://github.com/coq/coq/issues/17760

Thanks to @gadmm for all the interest and help.

Also cc: @HazardousPeach 

Bugs fixed when using the `memprof-limits` backend: #139 #484 #487

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,7 +332,7 @@ recommend that if you are installing via opam, you use the following branches
 that have some fixes backported:
 
 - For 8.20: No known problems
-- For 8.19: No known problems
+- For 8.19: `opam pin add coq-core https://github.com/ejgallego/coq.git#v8.19+lsp`
 - For 8.18: `opam pin add coq-core https://github.com/ejgallego/coq.git#v8.18+lsp`
 - For 8.17: `opam pin add coq-core https://github.com/ejgallego/coq.git#v8.17+lsp`
 - For 8.16: `opam pin add coq      https://github.com/ejgallego/coq.git#v8.16+lsp`
--- a/coq/limits.ml
+++ b/coq/limits.ml
@@ -57,8 +57,16 @@ let select = function
   | Coq -> backend := (module Coq)
   | Mp -> backend := (module Mp)
 
+(* Set this to false for 8.19 and lower *)
+let sane_coq_base_version = true
+
+let sane_coq_branch =
+  CString.string_contains ~where:Coq_config.version ~what:"+lsp"
+
+let safe_coq = sane_coq_base_version || sane_coq_branch
+
 let select_best = function
-  | None -> if Mp.available then select Mp else select Coq
+  | None -> if Mp.available && safe_coq then select Mp else select Coq
   | Some backend -> select backend
 
 module Token = struct
--- a/etc/doc/USER_MANUAL.md
+++ b/etc/doc/USER_MANUAL.md
@@ -79,7 +79,11 @@ not fully completed. Also, you can work with bullets and `coq-lsp`
 will automatically admit unfinished ones, so you can follow the
 natural proof structure.
 
-## Settings
+### Embedded Markdown and LaTeX documents
+
+
+
+## Coq LSP Settings
 
 ### Goal display
 
@@ -90,6 +94,9 @@ Several settings for goal display exist.
 A setting to have `coq-lsp` check documents continuously exists.
 
 ## Memory management
+
+You can tell the server to free up memory with the "Coq LSP: Free
+memory" command.
 
 ## Advanced: Multiple workspaces
 
@@ -119,10 +126,13 @@ on Coq <= 8.19 do need to install a version of Coq with the backported
 fixes. See the information about Coq upstream bugs in the README for
 more information about available branches.
 
+`coq-lsp` will reject to enable the new interruption mode by default
+on Coq < 8.20 unless the `lsp` Coq branch version is detected.
+
 ## Advanced incremental tricks
 
 You can use the `Reset $id` and `Back $steps` commands to isolate
-parts of the document from each others in terms of rechecking.
+parts of the document from each other in terms of rechecking.
 
 For example, the command `Reset $id` will make the parts of the
 document after it use the state before the node `id` was found. Thus,


ejgallego:
> We could solve this in OCaml 5 both with Domains or algebraic effects + continuations, however....

Actually I'm unsure that could work, indeed semantics of async effects are quite a bit more complex than semantics for regular ones, IIUC (c.f. https://arxiv.org/abs/2003.02110 ) .

@gadmm , is it feasible to think that async interruptions here (for token or limits) could be handled in OCaml 5 as an effect that could be resumed?

ejgallego:
I think this needs more testing, so I will postpone to 0.1.8 for now, as I'd like to release 0.1.7 ASAP.

gadmm:
If the question is about asynchronously pausing/restarting instead of immediately cancelling, you might be overthinking it as you can probably do this with (sys)threads using condition variables from an async callback (whether with signals & signal masks, or with an approach à la memprof-limits).

ejgallego:
> If the question is about asynchronously pausing/restarting instead of immediately cancelling

Yes, that's the question.

Usually we have a very slow `Qed`, which is busy with some large terms in conversion (so no polling available)

With `memprof-limits` I can now interrupt the slow `Qed`, which is good, however after the interruption, the manager may decide that it'd like to resume the `Qed` computation.

Let's assume for simplicity that all the computations here are pure (or properly guarded).

>  you might be overthinking it as you can probably do this with (sys)threads using condition variables from an async callback (whether with signals & signal masks, or with an approach à la memprof-limits).

Would you minding explaining a bit more how that would work in the above case?

I guess what you suggest is to have the async handler determine whether the computation should be interrupted or resumed, then act accordingly ? But I'm unsure what the limitations there would be, for example if the hanlder needs to do IO before resuming.

I was thinking of an interface more similar to effects, where the handler gets the continuation, which can be resumed or not at any later arbitrary point.

gadmm:
Using condition variables, you let the OS deal with scheduling. In this scenario there is no "async handler" and no need for one.

There is only an async callback, that is called inside the thread to be paused/interrupted. Inside this async callback, it is possible to wait on a condition variable as with control_check below:

```ocaml
type thread_status = Running | Paused | Interrupted

let status = { mutable status : thread_status; mutex : Mutex.create (); cond : Condition.create () }

(* from thread 1 *)
let control_set new_status =
  Mutex.lock status.mutex;
  status.status <- new_status;
  if new_status != Paused then Condition.broadcast status.cond;
  Mutex.unlock status.mutex

(* from an async callback in thread 2 *)
let control_check () =
  Mutex.lock status.mutex;
  while status.status = Paused do
    Condition.wait status.cond status.mutex
  done;
  let interrupted = (status.status = Interrupted) in
  Mutex.unlock status.mutex;
  if interrupted then raise Interrupted
```

Note that in your current approach, this involve changing memprof-limits itself to do something like that.

gadmm:
To really answer your question though:

> is it feasible to think that async interruptions here (for token or limits) could be handled in OCaml 5 as an effect that could be resumed?

Currently, it is not possible to perform an effect inside an async callback and handle it outside of the callback. There is no fundamental obstacle to implementing this, but: 1) this requires some work on the runtime to radically change how async callbacks run (essentially remove the need to switch from OCaml to C when running async actions), 2) you would need to convince upstream that this is something which they would like to have.

HazardousPeach:
Hey sorry, recently realized that I wasn't getting notified on github mentions. I'll test this right away, this seems super useful.

ejgallego:
Hi @HazardousPeach , we actually did a lot of progress with this implementation; I'll push a new tree soon, after discussion with Guillaume (the author of the lib we use here) we'd like to do some significant changes.

I'll be happy to merge this upstream once it gets a bit of testing.

Note that for your use case, we have finally worked (after your feedback and others) a protocol that should work much much better than LSP, but it will use the coq-lsp engine underneath.

HazardousPeach:
Currently failing to build with the error:
```
# File "coq/ast.ml", line 165, characters 53-55:
# 165 |     let children = List.map (projection_info ~lines) pi in
#                                                                                                  ^^
# Error: This expression has type
#          (Vernacexpr.local_decl_expr * Vernacexpr.record_field_attr_unparsed)
#          list
#        but an expression was expected of type
#          (Vernacexpr.local_decl_expr * Vernacexpr.record_field_attr) list
#        Type Vernacexpr.record_field_attr_unparsed is not compatible with type
#          Vernacexpr.record_field_attr
```

ejgallego:
Hi @HazardousPeach , this should be ready to try if you are curious.

HazardousPeach:
Huh, so I'm still not able to interrupt coq-lsp properly on this branch. If I send an interrupt signal (Ctrl-C) to the process, then it just quits the whole process, rather than stopping the current command. And if I just try to send a new command to be run instead of the old one, coq-lsp doesn't respond.

ejgallego:
> Huh, so I'm still not able to interrupt coq-lsp properly on this branch. If I send an interrupt signal (Ctrl-C) to the process, then it just quits the whole process, rather than stopping the current command. And if I just try to send a new command to be run instead of the old one, coq-lsp doesn't respond.

Thank for the feedback @HazardousPeach , just to be sure, have you passed the `--int_backend` parameter? What OCaml version are you on?

We didn't implement signal interruptions yet, so it is supposed to be interrupted when a new command arrives.

I will implement support for signal-based interruptions after I prepare everything to work fine in OCaml 5 multithreading, unless there is some urgency.

My tests with interrupting in a new command worked super well, so that's why I didn't bother so far.

HazardousPeach:
Oooh, I *hadn't* passed the --int_backend parameter, is that documented somewhere? I didn't see anything about it in this pull request thread or the docs, but when I looked at the files changed I saw it in the source code and found out that "Mp" is the value I should pass from there.

I'm on OCaml 5.1.1, is that new enough? The coq-lsp.opam file says anything greater than 5.0 is fine, but I know there have been problems with that file.

I'm still not really able to interrupt, but the behavior has changed slightly. After running a slow command, `compute`, I get a warning about a serlib plugin being missing (which I understand to be expected behavior), same as before. I can send a didChange request with the same document, but the `compute` (the last tactic) removed, and then request proof/goals. Whereas before this would just hang here, now I get *another* serlib plugin warning, but it still refers to the old document version. The proof/goals request isn't responded to though, and after that any didChange+proof/goals requests go unresponded to, even as they incrementally strip away every tactic in the proof in an attempt to backtrack to a good state.

ejgallego:
@HazardousPeach indeed we've merging the new interruption system gradually, full support is ready for the next release and already in `main`

We do indeed need to document this better before the release, in order to use the new interruption backend you _both_ need:
- OCaml 4.x: unfortunately the memprof-limits library only supports OCaml 4.x for now, it will take a bit more to port the new limits system to OCaml 5.x due to runtime changes
- to pass the `--int_backend` parameter to coq-lsp / fcc

The OCaml 4.x dep is unfortunate indeed, moreover when we almost reached a state where we can profit from OCaml 5.x new parallelism facilities.

HazardousPeach:
Ahh okay, I'll start recompiling everything for OCaml 4.x . Why does the coq-lsp.opam file say that only OCaml 5.x is supported? Even after the memprof stuff is ported to OCaml 5.x, some people might want to use coq-lsp with OCaml 4.x, right?

ejgallego:
@HazardousPeach coq-lsp opam file says:

```
  ("ocaml" {>= "5.0"} | ("ocaml" {<= "5.0"} & "memprof-limits" { >= "0.2.1" } ))
```
which means either ocaml >= 5.0 , or ocaml < 5.0 and memprof-limits, so we require only the library for the new limits when in OCaml 4.x [also I just spotted a bug, it should be `ocaml < 5.0` not `<=`

By the way, I just ported the serial part of Flèche to 8.11, not sure if that's still useful to you, but ping me in private if it would.

HazardousPeach:
Huh, so once I switched back to the 4.x OCaml compiler, I started getting the Prelude Not Found error again. To be sure, I created an opam switch from scratch with the following script, from inside the `coq-lsp` directory:
```
#!/usr/bin/env bash
set -e
opam switch create coq-lsp ocaml-base-compiler.4.14.2
eval $(opam env --switch=coq-lsp --set-switch)
opam install vendor/coq/coq{-core,-stdlib,ide-server,}.opam -y
opam install vendor/coq-serapi -y
opam install . -y
```
but the error persists, even though the same commands worked fine on OCaml 5.x.

Any ideas?

ejgallego:
@HazardousPeach is your `vendor/coq` repository clean?

opam has weird rules, so maybe running `git clean -fxd` in `vendor/coq` could help. It could also be that your opam version is old and has some pinning bugs / lack of features?

[Indeed opam has to correctly pick the updated opam file from `vendor` and only moderm opam do]

ejgallego:
Also check that the `coq-lsp.opam` file contains the `rm -r vendor` which is critical.

HazardousPeach:
My `vendor/coq` is clean. My Opam version is 2.1.2, is that recent enough? But I didn't see `rm -r vendor` in coq-lsp.opam, and realized that my code was a bit behind (I hadn't updated since you added that I guess), so I fetched the latest pull request code, deleted the switch, and recreated it with the same script. This failed the first time around because something wasn't right with how opam scheduled the package installs on the last command; the opam install of vendor/coq-core (among others)

ejgallego:
Oh indeed sorry, the problem with memprof-limits is annoying, I will update the instructions.

Indeed you need the latest code. Thanks a lot for the testing, I guess the next step is for me to try to reproduce the issue, would it be possible to share a problematic example?

Thanks again!

ejgallego:
@HazardousPeach note that examples in https://github.com/ejgallego/coq-lsp/issues/139 https://github.com/ejgallego/coq-lsp/issues/484 https://github.com/ejgallego/coq-lsp/issues/487 work correctly with the new interruption backend, so maybe these are good tests on your side.

The fix for Coq that I was waiting has been merged, so I will complete the docs, tweak display and usability and close this PR soon.

HazardousPeach:
Hmm, so the example from #487 doesn't interrupt properly on my end, so maybe there's a problem with how my coq-lsp is set up. Could you post the OCaml and memprof-limits versions that you've tested with? Could also be something about the way I'm driving coq-lsp in python versus through VSCode.

ejgallego:
@HazardousPeach , I have no problem interrupting the example from regular coq-lsp, so I cannot reproduce.

Will check serapy

ejgallego:
So indeed things don't work in SerAPY, trying to understand a bit more, however the way the code is structured now makes it pretty hard to debug (why is the code checking trace messages? Also, the way requests are handled seems strange)

ejgallego:
Hi @HazardousPeach , I did some more debug, and indeed, I cannot reproduce with the standard node LSP client.

You can find the example here: https://github.com/ejgallego/coq-lsp/pull/626

If you run the recommended command, the sequence serapy is trying works as expected.

I wouldn't be surprised if this is down to some low-level IO Python vs node differences, so IMHO the next step to debug is to be able to display what serapy is sending in the wire.

ejgallego:
I did some further checks, and actually the server is working fine; it seems that for some reason serapy gets deadlocked.

The trace for `test.py is:
```
=> "textDocument/didChange"  { [version: 20] "... unfold map. simpl."
=> "proof/goals" { id : 19 } { [version N/A] }
[process_queue]: Serving Request: textDocument/didChange
[process_queue]: Serving Request: proof/goals
[request]: postponing rq : 19
<= "textDocument/publishDiagnostics" { [version: 20], partial=true }
=> "textDocument/didChange"  { [version: 21] "... unfold map."
[process_queue]: Serving Request: textDocument/didChange
{"jsonrpc":"2.0","id":19,"error":{"code":-32802,"message":"Request got old in server"}}
[process_queue]: Serving Request: proof/goals
[request]: postponing rq : 20
<= "textDocument/publishDiagnostics" { [version: 21], partial=true }
<= "textDocument/publishDiagnostics" { [version: 21], partial=false }
[serving]: rq: 20 | {k:pos | l: 17, c: 11 v:  p: true}
=> "proof/goals" { id : 20 } { answer... }
```

and continues like this, with coq-lsp mostly idle between each `didChange` notification

HazardousPeach:
Ah, huh. When I did my own tracing I didn't see those `proof/goals` responses, maybe my setup has some difference causing an issue. If so, then I wouldn't have had a chance to experience the correct cancelling behavior, so my handling code would be untested and not ready for the correct behavior.

I've got some other tasks I have to catch up on, but I should be able to get back to this later in the week. When I do, I'll re-implement the tracing on my end and see if the behavior still differs, and I'll look into using `petanque` as a backend instead. It certainly seems like an easier api for what I'm doing, and the `ocaml-in-python` api would make it even more attractive (though the readme says it doesn't work yet).

However, the RL is already implemented fully on top of `coq_serapy` with the `coq-serapi` backend, the only issue is that backtracking is too expensive for smart search to work as well as it should in `coq-serapi` (but we're under submission anyway). The `coq-lsp` backend support was meant to fix this issue, and the support is complete except for interruption; the RL algorithm functions fully on proof searches where interruption is not necessary. So, I'll have to assess whether it makes sense to build a `petanque` backend right away, since the `coq-lsp` logic is already fully implemented (except for handling interrupt responses properly).

Relevant to this, interruption in `petanque` works the same as interruption in `coq-lsp`, right? So, if there's an issue with my setup causing problems interrupting `coq-lsp`, then that same issue would affect `petanque`?

ejgallego:
> Ah, huh. When I did my own tracing I didn't see those `proof/goals` responses, maybe my setup has some difference causing an issue. If so, then I wouldn't have had a chance to experience the correct cancelling behavior, so my handling code would be untested and not ready for the correct behavior.

So far I've done all my tests on `main` and worked fine, your backend also passed the parameters correctly.

I had to hack the server to realize it was working fine, tho you should see that in the test case you posted, CPU usage is 0%, however, when you pass `--int_backend=Coq`, it should go to 100% CPU use.

> It certainly seems like an easier api for what I'm doing

Yes, and it will be hugely more efficient. It's been on my mind to fix the problems you folks had on SerAPI for a while, but it took a long time to get the pre-requisites ready.

Now the main todo is multi-threading, I have a patch that could kind of work, however memprof-limits is not available in OCaml 5 :( However I think for RL-like applications, multi-threading in the server is a lower priority than for UI development.

> and the ocaml-in-python api would make it even more attractive (though the readme says it doesn't work yet).

That would be amazing indeed, however I didn't manage to get it working, must be some silly thing. I wonder if @thierry-martinez had a chance to look into these problems.

> I'll have to assess whether it makes sense to build a petanque backend right away, since the coq-lsp logic is already fully implemented (except for handling interrupt responses properly).

Indeed it is hard for me to asses what would make sense in your case, but I think the problems w.r.t. request will be similar both for petanque and LSP.

We have just considered using JSON-RPC as the official transport for petanque as a server, so in this case, it would share the request base with LSP. IMHO that's a very good thing.

In general, you need to handle all RPC calls as an async request, and check the error cases properly. JSON-RPC provides reasonable settings for this and seems popular enough.

> Relevant to this, interruption in petanque works the same as interruption in coq-lsp, right? So, if there's an issue with my setup causing problems interrupting coq-lsp, then that same issue would affect petanque?

I think the issue in your backend was not due to interruptions per-se but due to the cancellation of an async request. Note that this can happen in many cases, not only in interruptions (for example the server can inspect its queue and decide it got enough work to do).

So yes, the issue is very likely to occur in petanque.

Both `coq-lsp` and `petanque` are kind of flexible w.r.t. interruptions, for now, they are set to interrupt when a new command has been received, but could do otherwise, including having an expliciit `Interrupt` command.

ejgallego:
Note that Guillaume is coding a python backend for petanque, so maybe that could be plugged into your infra easily.

ejgallego:
Last bit of support has been added.

PR #517, opened 2023-06-28
[protocol] Rework verbosity into more granular options.

ejgallego:


ejgallego:
@HazardousPeach I have noticed that the logic for `verbosity = 1` and diagnostics is still inverted, are you sure this is the behavior you want.

The notification guarded by the flag is the one that takes time at document end. I guess you are instead relying on the `eager_diagnostics` flag set to true. Depending on your workflow that could be OK, but also could have a perf hit.

PR #537, opened 2023-08-10
[deps] Bump JS libs to 0.17

ejgallego:


Alizter:
What's the point of upper bounds actually?

ejgallego:
> What's the point of upper bounds actually?

Previous ones were needed because indeed serapi was not compatible with 0.17.

I prefer to keep them so CI is stable in the event for a 0.18 release.

ejgallego:
Note that indeed I'm not a fan of upper bounds, but in this case I feel they are justified due to historical problems with serlib and JS libs breaking often on JS version bump.

PR #544, opened 2023-09-15
chore: update flake

Alizter:


Alizter:
@ejgallego Looks like serlib isn't building?

ejgallego:
> @ejgallego Looks like serlib isn't building?

That seems pretty odd, I guess a problem with some newer ppx version?

I'll have a look when I can.

Alizter:
There are hardcoded versions in the flake which were causing the issue. I've pushed a fix, let's see if it works.

Alizter:
Looks like it worked. There was a TS formatting issue which I also fixed.

Alizter:
FTR you can easily update the format for nix, TS and ocaml by doing `nix fmt`.

PR #565, opened 2023-10-02
[client] Update vscode client and misc js packages

ejgallego:
Quite a few bug fixes that are convenient.

Not sure what happened with the tsconfig.json for the view.


Alizter:
We could consider using dependabot for updating these js deps.

ejgallego:
Yes; note tho that the `vscode` API required manual tweaking, some other manual actions were needed like the update of the tsconfig file. So I dunno if dependabot will get the udpates right.

Alizter:
@ejgallego It will just create PRs and keep open the ones that pass CI. The ones that fail will need intervention. AFAIK.

PR #573, opened 2023-10-10
[README] [nix] Add changes + tweak on the README.

ejgallego:
Follow up of #572

cc: @Zimmi48 @Alizter

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,6 +36,9 @@
  - New lexical qed detection error recovery rule; this makes a very
    large usability difference in practice when editing inside proofs.
    (@ejgallego, #567, fixes #33)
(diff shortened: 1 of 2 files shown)


Alizter (approved):

Zimmi48 (approved):

ejgallego:
Thanks folks; something I've been wondering is that we have Windows / MacOS / Linux install instructions, should Nix become another category, or should become a suboption in for the `MacOS / Linux` entries in the content table?

Alizter:
I think its fine to keep Nix as a separate category since Nix users will find it more easily and there really isn't any difference between macos and linux at that point.

ejgallego:
Ok, I think I'll add Nix to the start of the README too.

PR #576, opened 2023-10-13
[client] [vscode] Export Query Goals API

### LINKED ISSUE #558: [vscode client] graph-editor integration
I'm opening this issue to ensure the next release will have everything needed by @amblafont 's [graph-editor](https://github.com/amblafont/graph-editor-web)

@amblafont , do you have a pointer to the current hooks you are using? TIA!

IMHO, the thing that makes most sense is to place a Graph Editor extension in the Marketplace. Then, the extension registers with `coq-lsp`. But we don't need to do this at first.

Having a separate extension has some pros and cons:
- pros [in tree]: easy to install and to coordiante changes
- pros [separate]: own release schedule, more freedom in terms of organization, can support other Coq / Lean editors from the same extension.

See also #538 for the case of ViZX

ejgallego:
This way other extensions can implement their own commands that query Coq goals

With #574, close #558

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -42,6 +42,9 @@
  - Support for `-rifrom` in `_CoqProject` and in command line
    (`--rifrom`). Thanks to Lasse Blaauwbroek for the report.
    (@ejgallego, #581, fixes #579)
+ - Export Query Goals API in VSCode client; this way other extensions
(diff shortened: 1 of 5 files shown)


ejgallego:
cc @amblafont , let me know how this looks for you.

ejgallego:
We discussed offline and Ambroise valided the patch.

PR #590, opened 2023-10-26
[hover] [wip] First prototype of show documentation on hover

ejgallego:


Alizter:
Very cool! This only works with documentation in the same file though right? If we wanted to make it work across files we would need to get better location information of identifiers from Coq.

ejgallego:
> Very cool! This only works with documentation in the same file though right? If we wanted to make it work across files we would need to get better location information of identifiers from Coq.

Indeed, this will work for other files once we can have the per-theory `Toc.t`; the good news is that the code here doesn't need update, but still a lot to do to have coq-lsp working on full projects.

Issue #596, opened 2023-11-07
[vscode] Extension is not actived for `.mv` files


ejgallego:
Since 586dbbe1d28823858c4284dcaa078867026189a3 (cc #519) , it seems that the extension is not properly activated on `.mv` files. I have checked the commit with a bisect.

This bug was reported by @Zimmi48  (thanks!) , but I've noticed it as well.

I have opened a discussion thread upstream [1] , however I'm still unsure as to what is going on.

It seems to me that there is certainly something strange on this whole setup, I wouldn't discard a VSCode bug. With the current code, I see:
- if `coq-lsp` was previously activated, for example by opening a `.v` file, things work fine, and `.mv` files are activated
- however, `coq-lsp` won't auto-activate when only when using `.mv` files.

This makes little sense to me.

I cc @4ever2 as they are the original author of #519 and may know what is going on.

[1] https://github.com/microsoft/vscode-discussions/discussions/901

ejgallego:
Seemingly working workaround in #598 , but still not sure this is an upstream bug, keeping this open until we figure it out.

ejgallego:
Not an upstream bug, see https://github.com/microsoft/vscode/issues/198295

PR #598, opened 2023-11-07
[vscode] Workaround broken activation for `.mv` files

ejgallego:
Suggested by Tomer Chachamu (thanks!) in
https://github.com/microsoft/vscode-discussions/discussions/901

Should fix issue #596 , however I'm not marking this as closed, as this workaround could actually be due to a VSCode bug.

Thanks to Théo Zimmerman for the report.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,9 @@
  - fix hover position computation on the presence of Utf characters
    (@ejgallego, #597, thanks to Pierre Courtieu for the report and
    example, closes #594)
+ - workaround seemingly VSCode activation bug that prevented extension
+   activation (@ejgallego @r3m0t, #598, cc #596, reported by Théo
+   Zimmerman)
 
 # coq-lsp 0.1.8: Trick-or-treat
 -------------------------------
--- a/editor/code/package.json
+++ b/editor/code/package.json
@@ -28,6 +28,9 @@
     "type": "git",
     "url": "https://github.com/ejgallego/coq-lsp"
   },
+  "activationEvents": [
+    "onLanguage:markdown"
+  ],
   "contributes": {
     "languages": [
       {


CHANGES.md
```diff
@@ -10,6 +10,9 @@
  - fix hover position computation on the presence of Utf characters
    (@ejgallego, #597, thanks to Pierre Courtieu for the report and
    example, closes #594)
+ - workaround seemingly VSCode activation bug that prevented extension
+   activation (@ejgallego @r3m0t, #598, cc #596, reported by Théo
+   Zimmerman)
```
Zimmi48, line 15: You should probably edit the changelog entry if this is not a VS Code bug and the fix here is the proper fix as opposed to a workaround.

ejgallego: Indeed, thanks for pointing it out!

Zimmi48:
There is a serious issue though with pretending `mv` are just Markdown files and then associating coq-lsp with Markdown files: when opening a non-`mv` file (or even a Jupyter notebook with Markdown cells), the coq-lsp extension ends up being triggered and opening its Goals window. In my course where I use Coder, I have a first practical session, in the same VS Code workspace, which is not about Coq and uses a Jupyter notebook. I noticed that the Goals window kept appearing (even after being closed) when students interacted with the notebook (more specifically when interacting with the Markdown cells). I wonder if there isn't a solution that would e.g., add a new alias like CoqMarkdown for the `mv` files and then use this alias to activate the Coq extension...

ejgallego:
@Zimmi48 what VSCode version are you on?

ejgallego:
I did test for this particular problem before closing this issue, but indeed it seems I didn't test for this properly; I can observe the behavior you are pointing out.

I have prepared a fix at least for the Goals window issue, see #625 

Indeed the situation is maybe still not good? It seems to me upstream is still evolving activation rules; I'd be great if someone could gather more info; I'm afraid I won't be able to interact more with them in the short term.

PR #613, opened 2023-11-19
Follow LSP and treat positions as UTF16 code units

### LINKED ISSUE #531: No unicode completion after a mathcal character
**Describe the bug**

The `\` character does not trigger a unicode completion if there there is a mathcal character on the same line before it.

**To Reproduce**
1. Open a coq file
2. type 𝒰 using `\calU`
3. type on the same line
4. type `\`
5. The unicode menu does not open

**Expected behavior**
The unicode completion menu opens

**Desktop (please complete the following information):**
 - Operating system linux
 - coq-lsp version 0.1.7 (also 0.1.6)
 - VS Code version 1.80.1 (but also older versions from the spring)

ineol:
Fixes #531 

This is probably only a partial as it only affects completion requests.

The translation should probably be done in `lsp_core.ml`, in `do_position_request` for example. But then, what should we do when the position is out of bound? pass None to the implementation? or terminate directly the request?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,7 +61,7 @@ jobs:
       - name: 🐫🐪🐫 Get dependencies
         run: opam exec -- make opam-deps
 
-      - name: 🐛 Test coq-lsp
+      - name: 🐛 Special Windows Config [only on Win CI]
         if: matrix.os == 'windows-latest'
         run: opam exec -- make winconfig
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,6 +36,8 @@
    @ejgallego, cc: #614)
  - error recovery: Recognize `Defined` and `Admitted` in lex recovery
    (@ejgallego, #616)
+ - completion: correctly understand UTF-16 code points on completion
+   request (Léo Stefanesco, #613, fixes #531)
 
 # coq-lsp 0.1.8: Trick-or-treat
 -------------------------------
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ opam-switch:
 # Install opam deps
 .PHONY: opam-deps
 opam-deps:
-	opam install ./coq-lsp.opam -y --deps-only
+	opam install ./coq-lsp.opam -y --deps-only --with-test
 
 # Install opam deps
 .PHONY: opam-dev-deps
--- a/controller/rq_completion.ml
+++ b/controller/rq_completion.ml
@@ -79,8 +79,8 @@ let validate_line ~(doc : Fleche.Doc.t) ~line =
 let validate_position ~doc ~point =
   let line, char = point in
   Option.bind (validate_line ~doc ~line) (fun line ->
-      Option.bind (Coq.Utf8.index_of_char ~line ~char) (fun index ->
-          Some (String.get line index)))
+      let char = Coq.Utf8.get_byte_offset_from_utf16_pos line char in
+      Option.bind char (fun index -> Some (String.get line index)))
 
 let get_char_at_point ~(doc : Fleche.Doc.t) ~point =
   let line, char = point in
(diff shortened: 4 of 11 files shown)


ejgallego:
Thanks a lot @ineol , indeed we need to fix this at the core, I have a branch but didn't get enough time to finish it.

The changes needed are a bit deep, actually.

The convention is that `Lang.Point.t` positions are unicode chars, and `Coq.Pos.t` positions are Coq byte offsets.

Turns out this is wrong, `Lang.Point.t` should be protocol-level positions, in this case, not unicode points, but UTF-16 points. That's not hard to do, however it is a bit harder to be sure we do the conversion efficiently.

So we need to update all the conversion functions so they are more abstract (so we can still do Coq to several codepoints); newer versions of LSP do allow for the client/server comms to be in UTF-8.

I'll merge the workaround for now; and hopefully we get to fix the whole setup first.

ejgallego:
> But then, what should we do when the position is out of bound? pass None to the implementation?

I need to check the LSP specification carefully, but my impression is that if the client sends an out-of-bounds positions, the request should indeed fail with the error code "invalid argument".

As far as I can see, when that situation would happen, that'd always be a client bug, right?

ineol:
> So we need to update all the conversion functions so they are more abstract (so we can still do Coq to several codepoints); newer versions of LSP do allow for the client/server comms to be in UTF-8.

Note that VSCode does not implement this extension as far as I understand https://github.com/microsoft/vscode-languageserver-node/issues/1224

I think both emacs/eglot and neovim lsp clients support encoding negotiation.

ejgallego:
> > So we need to update all the conversion functions so they are more abstract (so we can still do Coq to several codepoints); newer versions of LSP do allow for the client/server comms to be in UTF-8.
> 
> Note that VSCode does not implement this extension as far as I understand [microsoft/vscode-languageserver-node#1224](https://github.com/microsoft/vscode-languageserver-node/issues/1224)
> 
> I think both emacs/eglot and neovim lsp clients support encoding negotiation.

Yes, I assume we will need to support the UTF-16 setup for a long time; I'd like to support encoding negotiation tho as for Coq using UTF-8 will be more efficient; maybe it doesn't matter too much tho, but I'm unsure, there are very very large Coq files out there.

ineol:
> > But then, what should we do when the position is out of bound? pass None to the implementation?
> 
> I need to check the LSP specification carefully, but my impression is that if the client sends an out-of-bounds positions, the request should indeed fail with the error code "invalid argument".
> 
> As far as I can see, when that situation would happen, that'd always be a client bug, right?

At least in the section [Position](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position) it says:

> If the character value is greater than the line length it defaults back to the line length.

ejgallego:
Hi @ineol , I did fix a few compile problems and pushed to your branch, still need to add a wrapper for the missing functions in older OCaml.

ineol:
> Hi @ineol , I did fix a few compile problems and pushed to your branch
thanks!

> still need to add a wrapper for the missing functions in older OCaml.
So should I copy-paste from Stdlib in a compatibility module until the minimum supported version is 4.14? they are both LGPL so it should be fine.

ineol:
It's done. I didn't manage to test with an old ocaml version but according to the doc it should be fine.

ejgallego:
Thanks! 

I will fix the two remaining CI errors and merge, then I'll propagate your fix to other parts, it was very useful!

ejgallego:
I've added a changelog, and placed the tests under `test`. Hopefully CI passes now.

I've also added an example that showcases other parts of the codebase where locations are incorrectly treated as UTF-8 (on hover, for diagnostics).

@ineol I was wondering if the kind of testing we are doing here wouldn't be better served by `ppx_inline_test` ?

I will for now keep the tests as they are, but something to keep in mind as we rework this code a bit.

PR #619, opened 2023-11-23
[plugin] Goal dump plugin.

ejgallego:
This is a simple plugin that will output the Ast and goals of a document. This is just a start, there are many tweaks that could be performed indeed.

cc: @gbdrt


ejgallego:
I'm reopening this plugin example as how to do this has become a F.A.Q.

PR #623, opened 2023-11-24
Use ppx_inline_test for unit tests

ineol:
As suggested by @ejgallego in #613, use inline tests instead of alcotest.


ejgallego (approved):
Amazing, thanks so much!

Nix CI seems to need a tweak, I'll have a look later, but please feel free to merge yourself if you find a solution first.

Alizter:
Nix is easy, just replace the alcotest in flake.nix with ppx_inline_test.

ejgallego:
> Nix is easy, just replace the alcotest in flake.nix with ppx_inline_test.

The PR seems to do that already, that's why I'm confused actually.

Alizter:
Need to add it here too https://github.com/ejgallego/coq-lsp/blob/008c6c207f6e177f79e72804579e7e0168152023/flake.nix#L69

ejgallego:
I see, we are not running the tests on Nix.

coq/dune
```diff
@@ -3,4 +3,6 @@
  (public_name coq-lsp.coq)
  ; Unfortunate we have to link the STM due to the LTAC plugin
  ; depending on it, we should fix this upstream
+ (inline_tests)
+ (preprocess (pps ppx_inline_test))
```
Alizter, line 7: This adds the test library as a dependency to the executable, which is probably not what you want. I think it would be better to have a separate test library like before.

Alizter:
We shouldn't be adding test libraries as build dependencies anyhow. They should live in a separate library as they currently do with alcotest.

coq/dune
ejgallego: Does it? I think dune handles this properly.

ejgallego:
> We shouldn't be adding test libraries as build dependencies anyhow. They should live in a separate library as they currently do with alcotest.

Why? The whole idea of inline tests is to have the code next to the libraries, it is true that some people use a policy of splitting test libs , but so far this is a matter of taste.

Myself I'm fine with the tests living next to the use.

Alizter:
@ejgallego Yes, but you are adding another build dependency just so you are aware.

ejgallego:
> @ejgallego Yes, but you are adding another build dependency just so you are aware.

Yes; as it was the way it was done, I suggest we keep it this way.

But indeed if we find a problem with the new setup, we can just do a follow-up PR and split the tests to their own file.

I like (it's a personal preference, of course) when tests (and proofs) are near their corresponding definitions.

But indeed, an extra dep is a non trivial cost to pay.

PR #655, opened 2024-03-19
Add request to vizcar extension

bhaktishh:
Added a call to a visualizer that needs to query goals using coq-lsp, mirroring behavior for vizx.

### DIFF
--- a/editor/code/src/goals.ts
+++ b/editor/code/src/goals.ts
@@ -93,11 +93,10 @@ export class InfoPanel {
     commands.executeCommand("vizx.lspRender", goals);
   }
 
-    requestVizCarDisplay(goals: GoalAnswer<PpString>) {
+  requestVizCarDisplay(goals: GoalAnswer<PpString>) {
     console.log(goals);
     commands.executeCommand("vizcar.lspRender", goals);
   }
-
 
   // notify the info panel that we found an error
   requestError(e: any) {
--- a/editor/code/src/goals.ts
+++ b/editor/code/src/goals.ts
@@ -93,6 +93,12 @@ export class InfoPanel {
     commands.executeCommand("vizx.lspRender", goals);
   }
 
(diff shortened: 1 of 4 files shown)


ejgallego (approved):
Thanks! Can you add an entry in the CHANGES file?

There seems to be two errors in CI:
- a typo on a method name (legitimate error it seems)
- some formatting problems (make ts-fmt should fix it)

Feel free to merge once ready.

editor/code/src/goals.ts
```diff
@@ -140,5 +154,10 @@ export class InfoPanel {
       console.log("vizx active in updateFromServer");
       this.sendVizxRequest(client, strCursor);
     }
+    let vizcar = extensions.getExtension("inQWIRE.vizcar");
+    if (vizcar?.isActive) {
+      console.log("vizcar active in updateFromServer");
+      this.sendVizCarRequest(client, strCursor);
+    }
```
ejgallego, line 161: As discussed, we can just keep a list of callbacks here, and add a method to register such a callback in client.ts.

The current code does:
```typescript
// Extension API type (note this doesn't live in `lib` as this is VSCode specific)
export interface CoqLspAPI {
  /**
   * Query goals from Coq
   * @param params goal request parameters
   */
  goalsRequest(params: GoalRequest): Promise<GoalAnswer<PpString>>;
}

export function activateCoqLSP(
  context: ExtensionContext,
  clientFactory: ClientFactoryType
): CoqLspAPI ...
```

So indeed, we already export a mechanism for extensions to do their own goal request.

For this use case, I guess we would like to add a `registerGoalsAction` function (and maybe de-register too)

The type of such function should be indeed a callback that takes goals (we could do better)

Also, it seems to me that the current code is doing the goals request twice, we should do it only once, by moving the hooks to the resolution of the regular `goalRequest` promise.

PR #660, opened 2024-03-22
[coq] Adapt `Coq.Protect` interface to incorporate an interruption token

ejgallego:


ejgallego:
Unfortunately it seems that `memprof-limits` is not packaged on Nix.

https://github.com/NixOS/nixpkgs/issues/300480

PR #664, opened 2024-04-04
[perf] [code] Fix Coq perf view display.

ejgallego:
It got broken in #503 as we didn't plug the notification.

We also had to implement better storage management.

We additionally add a tough test from
https://github.com/coq/coq/issues/17999 , by Gaëtan Gilbert.


test/compiler/long_file/run.t
```diff
@@ -0,0 +1,14 @@
+Degenerate performance case for the fcc compiler
+  $ export FCC_TEST=true
+
+We now compile the challenging file:
+  $ fcc --root . ./test_2.v
```
SkySkimmer, line 5: I may be nicer to regenerate the file here where you can run arbitrary shell rather than put an unreviewable 30k lines file in the git commit.

ejgallego: Good idea, done thanks!

PR #672, opened 2024-04-05
[vscode] Provide generic user goal request hook.

### LINKED ISSUE #538: [vscode client] Export an API for ViZX integration
As documented [here](https://code.visualstudio.com/api/references/vscode-api#extensions), it's possible for extensions to export an API for other extensions to use. I want to add an API for `coq-lsp` but I'm not sure what functions would make sense to include in the API. Are there thoughts on the same?

ejgallego:
This allows extensions to register without having to modify ours; it should subsume previous hooks for `VIXZ` and `VIZCAR` visualizers.

There are many more improvements we can do to this API.

Fixes #538

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -78,6 +78,9 @@
  - The `coq-lsp.pp_type` VSCode client option now takes effect
    immediately, no more need to restart the server to get different
    goal display formats (@ejgallego, #675)
+ - new public VSCode extension API so other extensions can perform
+   actions when the user request the goals (@ejgallego, @bhaktishh,
+   #672, fixes #538)
 
 # coq-lsp 0.1.8.1: Spring fix
 -----------------------------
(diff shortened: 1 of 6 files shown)


ejgallego:
@bhaktishh can you review this and see if that would work?

Extensions should be ported easily, basically they can recover the previous semantics by doingl:

````typescript
let hook = coqLSPAPI.onUserGoals((goals) => commands.executeCommand("vizx.lspRender", goals));
```
or any other code they may want.

`coqLSPAPI` can be obtained from the usual `extensions.getExtension('coq-lsp').export` field.

Don't forget to call `hook.dispose()` to remove the binding in the extension's `deactivate` method.

bhaktishh (commented):
This looks good to me -- I will make the changes and see if the ported change works.

ejgallego:
@bhaktishh merging this as I plan to release ASAP, I did some testing and it worked fine locally.

PR #679, opened 2024-04-15
[fleche] New `InjectRequire` plugin API

ejgallego:
This way plugins can specify a list of extra imports for each file.

This may compose poorly, so note that in the documentation.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -84,6 +84,9 @@
  - Support Visual Studio Live Share URIs better (`vsls://`), in
    particular don't try to display goals if the URI is VSLS one
    (@ejgallego, #676)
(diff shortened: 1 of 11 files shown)


Alizter:
Are these Coq plugins or coq-lsp plugins? I'm guessing the use-case is for the visualizer extensions to easily import the parts they need when working on a file?

ejgallego:
This is a Fleche plugin, so they work both for coq-lsp, jsCoq, `fcc`.

The use case here is more for running baselines so far, we have a plugin that will replace all proofs by some concrete text, but sometimes we need to inject some extra requires, for example in order to support Tactician, or `From Hammer Require Import Hammer`

This API is far from final form tho.

PR #684, opened 2024-04-26
chore: update nix flake

Alizter:
Looks like treefmt also got an update meaning we have to indent.

### DIFF
--- a/editor/code/package-lock.json
+++ b/editor/code/package-lock.json
@@ -35,6 +35,214 @@
         "vscode": "^1.82.0"
       }
     },
+    "node_modules/@azure/abort-controller": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@azure/abort-controller/-/abort-controller-1.1.0.tgz",
+      "integrity": "sha512-TrRLIoSQVzfAJX9H1JeFjzAoDGcoK1IYX1UImfceTZpsyYfWr09Ss1aHW1y5TrrR3iq6RZLBwJ3E24uwPhwahw==",
+      "dev": true,
+      "dependencies": {
+        "tslib": "^2.2.0"
+      },
+      "engines": {
+        "node": ">=12.0.0"
+      }
+    },
+    "node_modules/@azure/core-auth": {
+      "version": "1.7.2",
+      "resolved": "https://registry.npmjs.org/@azure/core-auth/-/core-auth-1.7.2.tgz",
+      "integrity": "sha512-Igm/S3fDYmnMq1uKS38Ae1/m37B3zigdlZw+kocwEhh5GjyKjPrXKO2J6rzpC1wAxrNil/jX9BJRqBshyjnF3g==",
+      "dev": true,
+      "dependencies": {
+        "@azure/abort-controller": "^2.0.0",
+        "@azure/core-util": "^1.1.0",
+        "tslib": "^2.6.2"
+      },
+      "engines": {
+        "node": ">=18.0.0"
+      }
+    },
+    "node_modules/@azure/core-auth/node_modules/@azure/abort-controller": {
+      "version": "2.1.2",
+      "resolved": "https://registry.npmjs.org/@azure/abort-controller/-/abort-controller-2.1.2.tgz",
+      "integrity": "sha512-nBrLsEWm4J2u5LpAPjxADTlq3trDgVZZXHNKabeXZtpq3d3AbN/KGO82R87rdDz5/lYB024rtEf10/q0urNgsA==",
+      "dev": true,
+      "dependencies": {
+        "tslib": "^2.6.2"
+      },
+      "engines": {
+        "node": ">=18.0.0"
+      }
+    },
+    "node_modules/@azure/core-client": {
+      "version": "1.9.2",
+      "resolved": "https://registry.npmjs.org/@azure/core-client/-/core-client-1.9.2.tgz",
+      "integrity": "sha512-kRdry/rav3fUKHl/aDLd/pDLcB+4pOFwPPTVEExuMyaI5r+JBbMWqRbCY1pn5BniDaU3lRxO9eaQ1AmSMehl/w==",
+      "dev": true,
+      "dependencies": {
+        "@azure/abort-controller": "^2.0.0",
+        "@azure/core-auth": "^1.4.0",
+        "@azure/core-rest-pipeline": "^1.9.1",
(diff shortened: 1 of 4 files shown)


ejgallego:
Thanks @Alizter , indeed let's update the format. How does `treefmt` choose the formatting version tho? It should use package.json right?

Alizter:
Yeah, I guess I'll have to update that too.

ejgallego:
What I mean is that it is strange the formatting has changed, as `treefmt` should read `package.json` and `.ocamlformat` to determine the version of the formatters.

Alizter:
I know that works for ocamlformat but I have no idea how .ts is formatted. I've updated npm and that has seemed to fix it. Should I do something more conservative?

ejgallego:
TS code is formatted using prettier, so indeed, this PR bumps prettier to a new release.

But the weird thing is that the nix Flake should not use the newer prettier version unless we bump package.lock.json

I'm fine updating the prettier version, but IMHO that should usually go on its own PR?

The package.lock update seems a bit too large for my taste, not sure why so much stuff is being installed?

editor/code/package-lock.json
```diff
@@ -424,92 +637,91 @@
...
+      "version": "15.7.12",
+      "resolved": "https://registry.npmjs.org/@types/prop-types/-/prop-types-15.7.12.tgz",
+      "integrity": "sha512-5zvhXYtRNRluoE/jAp4GVsSduVUzNWKkOZrCDBWYtE7biZywwdC2AcEzg+cSMLFRfVgeAFqpfNabiPjxFddV1Q==",
       "dev": true
     },
     "node_modules/@types/react": {
-      "version": "18.2.32",
-      "resolved": "https://registry.npmjs.org/@types/react/-/react-18.2.32.tgz",
-      "integrity": "sha512-F0FVIZQ1x5Gxy/VYJb7XcWvCcHR28Sjwt1dXLspdIatfPq1MVACfnBDwKe6ANLxQ64riIJooXClpUR6oxTiepg==",
+      "version": "18.3.0",
```
ejgallego, line 670: I see a lot of stuff is bumped to newer releases, IMHO we should do that in its own PR.

editor/code/package-lock.json
```diff
@@ -1375,14 +1871,15 @@
       }
     },
     "node_modules/minimatch": {
-      "version": "5.1.6",
-      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-5.1.6.tgz",
-      "integrity": "sha512-lKwV/1brpG6mBUFHtb7NUmtABCb2WZZmm2wNiOA5hAb8VdCS4B3dtMWyvcoViccwAW/COERjXLt0zP1zXUN26g==",
+      "version": "3.1.2",
```
ejgallego, line 1874: This downgrade is weird

editor/code/package-lock.json
```diff
@@ -1931,9 +2448,9 @@
       }
     },
     "node_modules/typescript": {
-      "version": "5.2.2",
-      "resolved": "https://registry.npmjs.org/typescript/-/typescript-5.2.2.tgz",
-      "integrity": "sha512-mI4WrpHsbCIcwT9cF4FZvr80QUeKvsUsUvKDoR+X/7XHQH98xYD8YHZg7ANtz2GtZt/CBq2QJ0thkGJMHfqc1w==",
+      "version": "5.4.5",
```
ejgallego, line 2451: Another major upgrade here.

Alizter:
@ejgallego Can you update the formatter in it's own PR? I'm not really sure what I'm doing here. I'll rebase the flake update on that afterwards.

editor/code/src/client.ts
```diff
@@ -242,10 +242,10 @@ export function activateCoqLSP(
...
         ? 1
         : callKind == TextEditorSelectionChangeKind.Keyboard
-        ? 2
-        : callKind
-        ? callKind
-        : 3;
+          ? 2
+          : callKind
+            ? callKind
+            : 3;
```
ejgallego, line 248: Looks better yes!

Alizter:
I just wanted to have a more up-to-date development flake, I don't particularly care about the formatting details.

ejgallego:
> @ejgallego Can you update the formatter in it's own PR? I'm not really sure what I'm doing here. I'll rebase the flake update on that afterwards.

Sure, I'm happy to do so; however I'm confused how the Nix stuff determines the version to use in `prettier`.

Can we figure that out first?

Because if `treefmt` doesn't pick the prettier version from package.json we will indeed have problems with a PR updating prettier too.

Indeed, I'm not sure `treefmt` can support non-Nix development use cases.

ejgallego:
> I just wanted to have a more up-to-date development flake, I don't particularly care about the formatting details.

I can see that, but the flake needs to respect the `prettier` version set in package.json.lock , otherwise we will find trouble.

ejgallego:
So it seems the setting should be something like `programs.prettier.package = $(read_version_in_package.lock.json)` IIUC `treefmt` documentation properly.

ejgallego:
Shell code to find the version is:
```
(cd editor/code/ && npm ls --json --package-lock-only | jq '.dependencies.prettier.version')
```

ejgallego:
Another way: `npm pkg get devDependencies.prettier`

Anyways, I think you can just set `programs.prettier.package` to 3.0.3 and forget about it for now.

Alizter:
Supposedly treefmt is going to take care of the version. I'm going to experiment and see why it isn't adhering to it in this case. It could be that I bumped the flake to an unstable version of treefmt-nix or something.

ejgallego:
It seems to me that it is the norm in Nix to have package versions fixed in the hashes, so indeed, it is likely picking the wrong version.

ejgallego:
I think `nix-treefmt` doesn't do any special handling for prettier versions, but I could be wrong, I guess looking at the source code would say.

`treefmt` itself seems to rely on `npm i` doing the right thing.

[So that's our pain point, `npm i` and the flake not being in sync]

Alizter:
https://github.com/prettier/prettier/issues/7940

sigh...

I guess we will have to work out a way to pin this.

ejgallego:
What's wrong with setting `program.prettier.package=` in the flake?

Alizter:
@ejgallego That just let's us set the derivation, but the only one we have is the latest available in nixpkgs where they have tweaked how things are formatted. I can't see a way to set the version in here.

Issue #681, opened 2024-04-19
Heatmap for timing and other data


Alizter:
We have various bits of information available to the user via hover such as timing, memory usage, universe count etc.

It would be useful if the VSCode extension had a way of visualizing this data using a heatmap.

Here are some examples of extensions with heatmap functionality:
- https://github.com/SalmonMode/activity-heatmap
- https://github.com/chrisjdavies/vscode-heatmap

PR #686, opened 2024-04-28
WIP heatmap display for timing data

Alizter:
Here is a proof-of-concept for #681. Currently we generate random data and display that instead. I haven't been able to workout how we can get timing data yet.

### DIFF
--- a/editor/code/src/heatmap.ts
+++ b/editor/code/src/heatmap.ts
@@ -1,0 +1,115 @@
+import {
+  window,
+  workspace,
+  TextEditorDecorationType,
+  Range,
+  Selection,
+  Disposable,
+  TextEditor,
+  TextDocument,
+  languages
+} from "vscode";
+import { CoqSelector } from "./config";
+
+export class HeatMap {
+    private subscriptions: Disposable[] = [];
+    enabled: boolean = false;
+    heatStyles: TextEditorDecorationType[] = [];
+    
+    dispose() {
+        if (this.subscriptions) {
+            this.subscriptions.forEach(subscription => subscription.dispose());
+        }
+        this.heatStyles.forEach(style => style.dispose());
+    }
+
+    constructor(enabled: boolean) {
+        this.enabled = enabled;
+        this.activate();
+        this.registerEventListeners();
+    }
+    
+    private registerEventListeners() {
+        // Redraw when the active text editor changes
+        this.subscriptions.push(window.onDidChangeActiveTextEditor(editor => {
+            this.draw(editor);
+        }));
+
+        // Redraw when the document is edited
+        this.subscriptions.push(window.onDidChangeTextEditorSelection(event => {
+            this.draw(event.textEditor);
+        }));
+    }
+
+    toggle() {
+        this.enabled = !this.enabled;
+        if (this.enabled) {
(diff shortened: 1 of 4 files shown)


ejgallego:
Getting the data should not be hard, questions:
- does the client pull for the data, or should the server notify it
- how often does the data need to be refreshed?
- is the data incremental
- can we have an interface / OCaml record for the data

Alizter:
I think if we want it to be incremental the server should notify the client. Presumably when we do the progress updates.

Right now the data is refreshed everytime you move the cursor or activate the window which obviously doesn't scale too well. I would prefer that this is refreshed when a notification is recieved. The colouring is relative, so if we get a line with a large value relative to the rest, all the lines will need to be recoloured.

editor/code/src/heatmap.ts
```diff
@@ -0,0 +1,147 @@
...
+      // const startPos = document.positionAt(perf.loc.start);
+      // const endPos = document.positionAt(perf.loc.end);
+      const startPos = new Position(
+        perf.loc.line_nb - 1,
+        perf.loc.bp - perf.loc.bol_pos - 1
+      );
+      const endPos = new Position(
+        perf.loc.line_nb_last - 1,
+        perf.loc.ep - perf.loc.bol_pos_last - 1
+      );
```
ejgallego, line 100: Sorry this should be `Range` instead of Coq-style locs, will fix ASAP.

Alizter:
image

ejgallego:
Ok, rebased and working now!

IMHO I would cleanup this a bit, add an example, and maybe merge. I guess the heatmap will require some fine tuning with real files, but even in basic form that's a very cool addition.

Alizter:
This appears to be working really well. The configuration options also appear to work, though it is very likely there are some bugs. I set it up so that the options update the heatmap in real time without having the reload the extension which might be useful.

I think I want to add a hover functionality in heatmap mode so that we can actually see the values being highlighted. We obviously have something via the debug stats, but having something more user facing might be useful.

ejgallego:
Indeed, once the bugs in the backend were fixed, this seems to be pretty nice and not too complex!

> I think I want to add a hover functionality in heatmap mode so that we can actually see the values being highlighted. We obviously have something via the debug stats, but having something more user facing might be useful.

Maybe we could re-use the existing hover for performance data, in order to do this, we'd need:
- connect real time update of server options to server (I don't see any trouble with this)
- improve / rework the current hover for timing data so it is ready for user consumption
- have the heatmap toogle enable the server-side timing on hover plugin (careful with the logic here, as we need to restore the previous setting by the user)

Alizter:
@ejgallego I think this is ready.

ejgallego (approved):
LGTM, thanks Ali!

A couple of notes:
- as of now, we always update the CHANGES.md file, and on release we update the client changelog (this is described in contributing.md). Please move the changelog to the main CHANGES.md file, and if you wanna keep the client option, put it under `unreleased`, not under 0.1.8.
- please update the PR title as it still says "WIP"

ejgallego:
Maybe also squash if you want.

Alizter:
@ejgallego Changelogs updated and commits squashed.

PR #689, opened 2024-04-29
[perf] Fixes and improvements towards more interesting uses.

ejgallego:
Motivated by the nice heatmap (PR #686) , we fix and improve perf data, in particular, we

- we now send data for all the sentences, not only for the top 10
- we send data that takes into account the cached execution of the command
- we include a client-side option to disable sending of this data, as it can be "large" for large files (not so much actually, but just in case)
- we fix types that were wrong due to the mistakes in #503 / #513 , and weren't fully solved by #664

@Alizter , the PR is still not finished (we only report "real" time when a sentence was cached), but it should provide you a much better basis than `main` for your work on #686 . In particular, we now send perf data for all the sentences.

Let me know of any question.

TODO:

- [ ] Add client-side option to enable perf data
- [ ] Fix data creation in doc.ml
- [ ] Fix stored data validation in Perf view (maybe use a timestamp?)
- [ ] cleanup rest of XXX


editor/code/lib/types.ts
```diff
@@ -113,12 +114,22 @@ export interface FlecheSaveParams {
...
+  // Whether the execution was cached
+  cached: boolean;
+  // "Real" Execution Time, often very low for memoized sentences
+  real: number;
+  // Difference in words allocated in the heap using `Gc.quick_stat`
   mem: number;
 }
 
+export interface SentencePerfParams {
+  range: Range;
```
Alizter, line 129: This is not the correct range I believe. It should be vscode.Range.

ejgallego: This is part of the protocol so it should use the protocol types no? What would other clients do?

On top of that vscode ranges may not properly serialize

Alizter: Alright that makes sense. I did the other change in my PR, but I didn't realise that the serializability was an issue. There should be a standard conversion function between the two types right? I couldn't find it.

PR #698, opened 2024-05-09
[build] [deps] Merge `serlib` into coq-lsp

ejgallego:
It makes a lot more sense for us to develop `serlib` natively here, as we need new exclusive functionalities and the submodules are not bringing us a lot.

I guess we will reverse the dependency and have `sertop` depend on `coq-lsp` serlib's, if that's still needed.

TODO:
- [x] fix `genargs` tests
- [ ] update Coq's CI
- [x] update SerAPI's code to avoid duplication
  + [x] see https://github.com/ejgallego/coq-lsp/pull/698
  + [x] remove sertop_loader
  + [x] remove serapi_goals [decided against]
- [ ] update jsCoq


Alizter:
What happens to the opam package? Do you just switch the source?

ejgallego:
> What happens to the opam package? Do you just switch the source?

`coq-lsp` wont' depend anymore on `coq-serapi`

Newer `coq-serapi` will depend on `coq-lsp`, if there is a new release tho, likely the last.

PR #699, opened 2024-05-09
[vscode] cleanup command names

Alizter:


ejgallego (commented):
Ok to me, I'm not sure if some information on what the commands do is lost tho, hard for me to judge. Couple of points:

- coq-lsp.document: Maybe we should specify what format the document is, anyways this is just a debug call (unless we connect this call to a JSON visualizer, that could be cool actually! Maybe we could just put the output of the request in an ephemeral document-version.json file that is then opened?
- sentenceNext/Back: note that these commands are just an approximation, so they can often fail or be imprecise.
- trim: no idea if that sound too technical

Alizter:
The main reason for keeping these names short is that they show up on `Ctrl + P` and `>`. There isn't much room in that menu for long command descriptions.

Alizter:
For `trim` we could replace it with `Coq LSP: Free memory`

Alizter:
I'll rebase

Alizter:
@ejgallego I've rebased and tweaked the wording.

ejgallego:
Thanks Ali, I will tweak the working to say "serialize and display in a new window", as the command now opens a new windows.

PR #701, opened 2024-05-09
[code] Display JSON data for `coq-lsp.document` in new editor.

ejgallego:
This makes more sense than the current command behavior which is not usable by almost no one.

### DIFF
--- a/editor/code/package.json
+++ b/editor/code/package.json
@@ -96,7 +96,7 @@
       },
       {
         "command": "coq-lsp.document",
-        "title": "Coq LSP: Get document in serialized format, print in console"
+        "title": "Coq LSP: Display current Coq document in JSON format"
       },
       {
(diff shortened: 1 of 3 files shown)


editor/code/package.json
```diff
@@ -96,7 +96,7 @@
       },
       {
         "command": "coq-lsp.document",
-        "title": "Coq LSP: Get document in serialized format, print in console"
+        "title": "Coq LSP: Get document data in serialized JSON format, display in new text editor"
```
Alizter, line 99:
```suggestion
        "title": "Coq LSP: Serialize document to JSON"
```

ejgallego: Actually the main verb here now is "Display"

ejgallego: Reworked a bit, we can tweak further in #699

PR #702, opened 2024-05-10
[lsp] Update server settings on the fly.

ejgallego:
This means that we can finally tweak server options on the fly!

We use the `workspace/didChangeConfiguration` notification, however note that our use is still very rudimentary (and push-based) when compared to the recommended pull model using the
`workspace/configuration` request.


ejgallego:
@Alizter after this you can just update a hover setting normally, and things should work automatically.

PR #707, opened 2024-05-14
[ci] Test OCaml 5.2 series.

ejgallego:


Alizter:
There is now GC compaction in 5.2 IIUC. That might be useful. What is the status of OCaml 5 support for Coq? Is it worth switching over as a default version yet?

ejgallego:
> What is the status of OCaml 5 support for Coq? 

Mixed, it works better for some users, worse for others.

> Is it worth switching over as a default version yet?

Not yet, as `memprof-limits` is not available.

PR #708, opened 2024-05-14
[code] Correctly dispose even if client shutdown fails.

ejgallego:
This was likely causing some problems on extension restart.

Still we need to improve how we manage lifetime events.


Alizter:
I've ran into all sorts of bugs with the extension hanging around, but I typically just reload vscode in those cases which nukes any running instances. Unfortunately, I wasn't able to find anything reproducible to report a bug. Hopefully this improves the situation, but it isn't a critical issue.

ejgallego:
Yes, there are some lifetime problems. This will help in many cases actually!

Another important bug is on extension deactivation, we are doing `context.subscriptions.push` of object that actually have server-bound lifetime, not extension bound!

PR #709, opened 2024-05-15
[fleche] Support Coq Meta Commands Reset and Back

ejgallego:
They are pretty easy to support in our model, and actually I can see how they pretty useful to hint the incremental engine to ignore changes!

For example, see this document:
```coq

(* Coq code 1 ... *)

Lemma foo : ....

(* Coq code 2 ... *)

Reset foo.

(* Coq code 3 ... *)
```

In this case, `Reset` allows us to put a barrier so we can update Coq code 2 at will, and still don't re-check the 3rd code block as it depends on a previous state.


ejgallego:
This is ready.

The more I use this, the more I like it. Incrementality gives the commands a new meaning. For example, `Reset Initial` can be used to have multiple documents in one, all independent.

PR #718, opened 2024-05-17
[code] Add keybindings for sentenceNext / sentencePrevious

ejgallego:
`Alt+N` / `Alt+P`

### DIFF
--- a/editor/code/package.json
+++ b/editor/code/package.json
@@ -125,6 +125,18 @@
         "key": "alt+enter",
         "mac": "meta+enter",
         "when": "editorTextFocus && (editorLangId == coq || editorTextFocus && editorLangId == markdown)"
+      },
+      {
+        "command": "coq-lsp.sentenceNext",
+        "key": "Alt+N",
+        "mac": "cmd+N",
+        "when": "editorTextFocus && (editorLangId == coq || editorTextFocus && editorLangId == markdown)"
+      },
+      {
+        "command": "coq-lsp.sentenceBack",
+        "key": "Alt+P",
+        "mac": "cmd+P",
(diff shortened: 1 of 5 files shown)


Alizter:
Alternatively, we could rename sentenceBack to sentencePrev and keep the key binding. That would actually sound more correct.

ejgallego:
Yeah, Alt-P (previous) and Alt-N (next) are standard Emacs bindings, so I dunno. For experienced Coq users this make make sense due to ProofGeneral being the most used Coq IDE, but I dunno for others.

I propose we stick to `Alt-N/P` for now as they have much better hand ergonomy than `Alt-B`

ejgallego:
> Alternatively, we could rename sentenceBack to sentencePrev and keep the key binding. That would actually sound more correct.

Ok let's do that.

editor/code/package.json
```diff
@@ -111,7 +111,7 @@
         "title": "Coq LSP: try to jump to next Coq sentence"
       },
       {
-        "command": "coq-lsp.sentenceBack",
+        "command": "coq-lsp.sentencePrevious",
         "title": "Coq LSP: try to jump to previous Coq sentence"
```
Alizter, line 115: Also whilst we are here, let's change these to `Try` same here and above.

Alizter: Or alternatively drop the try and just start with `Jump`.

ejgallego: The thing is that the command really "tries" to jump, it can fail often.

ejgallego: Let's continue discussion on #699 , I'll be merge it before next release.

Issue #24, opened 2022-07-06
Lazy checking of documents


ejgallego:

ejgallego:
This requires parsing resumption so bumping milestone.

PR #721, opened 2024-05-19
[code] Display and allow status bar item to toggle continuous / lazy …

ejgallego:
…mode

We display the continuous / lazy mode checking in the status bar item, and we allow to toggle them with a click.

There are some usability questions with the current setup tho, I wonder if we should move the start / stop server command to the tooltip or to a `LanguageStatusItem` as done in other extensions.

cc: #24

### DIFF
--- a/editor/code/src/client.ts
+++ b/editor/code/src/client.ts
@@ -225,7 +225,7 @@ export function activateCoqLSP(
       .catch((error) => {
         let emsg = error.toString();
         console.log(`Error in coq-lsp start: ${emsg}`);
-        setFailedStatuBar(emsg);
+        setFailedStatusBar(emsg);
       });
   };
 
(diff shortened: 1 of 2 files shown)


editor/code/src/client.ts
```diff
@@ -431,9 +442,15 @@ export function activateCoqLSP(
...
     if (client && client.isRunning()) {
-      lspStatusItem.text = "$(check) coq-lsp (running)";
-      lspStatusItem.backgroundColor = undefined;
-      lspStatusItem.tooltip = "coq-lsp is running. Click to disable.";
+      if (serverConfig.check_only_on_request) {
+        lspStatusItem.text = "$(check) coq-lsp (on-demand checking)";
+        lspStatusItem.backgroundColor = undefined;
+        lspStatusItem.tooltip = "coq-lsp is running. Click to disable.";
+      } else {
+        lspStatusItem.text = "$(check) coq-lsp (continous checking)";
```
Alizter, line 450: I think this button should only appear when the language server is enabled via the other button.

ejgallego: What is the "other button"?

But yes, I think we should move this to its own language server status

Alizter: The other button is the enable/disable button that we already have. It doesn't make sense for the coq-lsp to be disabled and this button visible, IMO.

ejgallego: This is the same button, isn't it?

Alizter: Oh then I got confused. I thought you added another button when I read it this morning. Nevermind.

ejgallego: For now we have a single button with 4 states (enabled (continous/lazy) / disabled / failed), but I think we should move this stuff to a languageStatusItem actually, I will implement a `serverStatus` notification and do this, will look nice actually.

PR #722, opened 2024-05-22
Give focus back to Goals panel when navigating proofs

Alidra:
The `Goals` panel is shown in the second column of Vscode. If the user opens a second panel in this column the `Goals` panel may loose focus and proofs and not visible to user when he navigates the proofs again.

This PR, gives the focus back automatically to the `Goals` panel if it has lost it as soon as the user navigates the proofs again.

### DIFF
--- a/editor/code/CHANGELOG.md
+++ b/editor/code/CHANGELOG.md
@@ -1,6 +1,9 @@
 # coq-lsp 0.1.8: Trick-or-treat
 -------------------------------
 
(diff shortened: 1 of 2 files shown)


ejgallego (approved):
Thanks a lot @Alidra !

Would you mind adding a changelog?

[Also we recommend signing the commit as a way to id contributions]

Alidra:
You are welcome @ejgallego 
Done!

PR #727, opened 2024-05-28
[fleche] Support .tex literate documents

ejgallego:
Similarly to `.mv` files, we recognize `.lv` files to be LaTeX files with embedded Coq snippets, delimited by `begin/end` pairs.

Some tweaks will be likely needed.

```
\begin{coq}
\end{coq}
```


Alizter:
Why not just recognize .tex files?

ejgallego:
> Why not just recognize .tex files?

Seems a bit heavy or intrusive to me to start the coq-lsp server for all .tex files, but indeed I'm open to do this based on user feedback.

Alizter:
Right, I see why you wouldn't want that now. I'm thinking that latex tools won't work well with other file extensions. They generally all expect to see .tex. But as long as this is configurable this should be ok.

ejgallego:
> Right, I see why you wouldn't want that now. I'm thinking that latex tools won't work well with other file extensions. They generally all expect to see .tex. But as long as this is configurable this should be ok.

Yes, maybe we can change this to `.v.tex`  for now?

ejgallego:
Indeed I had already trouble using `.lv` but I like that idea, so I'll update the PR so both `.lv` and `.v.tex` work.

Issue #652, opened 2024-02-26
Improvement for goal display


qcfu-bu:
I have been running into an issue with the goal display of the vscode client that's a bit annoying. The display for big case analysis does not collapse when entering bullets. A toggle option to eagerly collapse bullets would be great.

<img width="2560" alt="Screenshot 2024-02-26 at 9 46 01 AM" src="https://github.com/ejgallego/coq-lsp/assets/56097784/b1f5e33c-8f57-4797-98b3-0c122eb24fee">

Moreover, the page always scroll down to prioritize displaying the goals on bottom instead of the one on the top. For most use cases, displaying the context would be more helpful.

<img width="2560" alt="Screenshot 2024-02-26 at 9 42 41 AM" src="https://github.com/ejgallego/coq-lsp/assets/56097784/bd7f251f-5174-4dee-ac0b-bd54db1c9c47">

qcfu-bu:
Additionally, while the default printer with syntax highlighting looks great, it is not very responsive when displaying a long list of goals. The string and layout-engine printers are much faster but don't look as nice.

Alizter:
I am also frequently bitten by this issue and I kept meaning to open an issue about it. Thanks for doing so. I would suggest that we collapse goals by default and make it a configurable behaviour as I think mathcomp routinely work on multiple goals at the same time, so the user should have control over the default behaviour.

I opened a similar issue, but about larger goals https://github.com/ejgallego/coq-lsp/issues/485. Additionally, we should have a way to collapse hypotheses too, but I suppose that's another issue.

Issue #525, opened 2023-07-07
[goals] Option to hide collapsed goal type display


ejgallego:
Indeed in some cases the density can become too high, requested by Anton Podkopaev (thanks!)

bhaktishh:
Can you elaborate a bit on what this is aiming for exactly? I'm working on something that i think might be similar and planning on having it being closely integrated w coq-lsp (it's mostly in the planning stage right now but happy to chat if you're interested)

Alizter:
Here is a screenshot demonstrating how annoying this issue is:
![image](https://github.com/ejgallego/coq-lsp/assets/8614547/b041e7ea-9425-4eaf-bae0-bad7dd40bc7d)

ejgallego:
Yes I agree this is messy; adding an option should be super easy tho, adding a UI button to hide the goals shouldn't be too hard to.

Alizter:
@ejgallego 

An example:
```coq
From Coq Require Import Prelude.
Inductive foo := a | b | c | d | e.

Goal forall x y z w v : foo, Type.
intros [].
- intros [].
  + intros [].
    * intros [].
      -- intros [].
      
```

Go down to the last bullet. Everything is expanded. Minimize everything. If you go up with the cursor and go down again everything is expanded again. I think the default behaviour should be minimisation.

PR #730, opened 2024-05-28
[code] Don't show goals type when details are closed.

ejgallego:
This is a workaround for #525 #652

For now, I disable this setting, so types are not displayed.

The right fix would be to add a checkbox in the view, so this can be toggled dynamically, but I don't have the cycles for that right now, and not worth delaying the release which is long overdue.

### DIFF
--- a/editor/code/views/info/Goals.tsx
+++ b/editor/code/views/info/Goals.tsx
@@ -61,16 +61,24 @@ function Goal({ goal, idx, open }: GoalP) {
     }
   });
 
+  // XXX: We want to add an option for this that can be set interactively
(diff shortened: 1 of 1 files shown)


Alizter:
Can you also tweak the unfocused goals to not be expanded? I believe it requires toggling some booleans but I had trouble working out which ones.

ejgallego:
I don't see what you mean: all the subgoals look like not expanded to me.

Do we have an example on `examples/goals.v` that I could look at?

ejgallego:
Will merge this meanwhile.

PR #735, opened 2024-05-30
[petanque] Return extra premise information.

ejgallego:
We now return the file where the premise is defined, as well as its position and text.


ejgallego:
Good as a first step.

PR #641, opened 2024-01-19
[wip] File auto-build

ejgallego:
This is joint work with @bhaktishh , opening this draft PR to coordinate the presentation of the OCaml 4.x prototype, and eventually merge into a proper release.

PR #744, opened 2024-06-03
[fleche] Caching / sharing of .vo contents

ejgallego:
We add a basic .vo parsing cache; this often results in large memory savings when working simultaneously with several files.

Based on work from #641 and coq/coq#17393

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,9 @@
    heuristic is enabled (@ejgallego, #748)
  - [memo] More precise hashing for Coq states, this improves cache
    performance quite a bit (@ejgallego, #751)
+ - [fleche] This achieves an almost 50% memory reduction for example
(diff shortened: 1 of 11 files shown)


ejgallego:
@Alizter memory use (approx) with patch, opening a lot of files in HoTT:

- with cache=true: 84Mw
- with cache=false: 144Mw

So while not the panacea, the reduction in memory use seems quite useful, in particular for devs with large .vo size.

ejgallego:
cc: @SkySkimmer @ppedrot in case you are curious about the kind of improvements we see by sharing .vo files (in stdlib that amounts to 8% faster compilation)

ejgallego:
Loading all of HoTT in the same server: 1169 Mw to 668 Mw , quite a considerable saving @Alizter

PR #754, opened 2024-06-05
Fix help printing

ineol:
Currently, `coq-lsp --help` crashed because `Cmdliner` cannot print the default option for `--int-backend`

It would be nicer to have `help` print the possible values of the enum, but I couldn't figure out how.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,9 @@
    when opening all of HoTT .v files (@ejgallego, @SkySkimmer,
    @bhaktishh, #744)
  - [memo] Provide API to query Hashtbl stats (@ejgallego, #753)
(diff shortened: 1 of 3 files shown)


ejgallego:
Oh! What a silly bug!

Thanks a lot for the fix.

There is also this option, printing the values as you suggest. It also avoids the dummy `auto` argument.
```ocaml
  let backends = [ ("Coq", Limits.Coq); ("Mp", Limits.Mp) ] in
  let docv = Cmdliner.Arg.doc_alts_enum ~quoted:true backends in
  Arg.(
    value
    & opt (some (enum backends)) None
    & info [ "int_backend" ] ~docv ~doc)
```

PR #766, opened 2024-06-07
[petanque] Allow `init` to be called multiple times

ejgallego:
We need to rework the protocol as to better reflect the Coq/Flèche workspace mode, but this hotfix will work meanwhile.


ejgallego:
cc @gbdrt

gbdrt (approved):
Thanks @ejgallego! That's perfect.
Just checked with pytanque and it works as expected.

PR #782, opened 2024-06-12
[petanque] Always initialize default workspace.

ejgallego:
This fixes a bug if `pet-server` was initialized without `--root`


ejgallego:
@gbdrt let me know if that improves things; will look into the problem with premises.

Note that setting `--root` to a non-absolute path is not really supported, if this is the problem with premises then it will unfortunately take a while to fix.

PR #783, opened 2024-06-12
[petanque] New methods `state/eq` and `state/hash`

ejgallego:
This allows the client to implement a state hashtable; see the
comments on the API for low-level details.

Note the changes in `run` result type. I have actually slightly refactored the type; @gbdrt please let me know if that's OK for you, or you would prefer the older type?


ejgallego:
@gbdrt also please keep an eye on collisions once using this; I tuned the hashing function for document states, these could behave pretty different than proof states.

PR #813, opened 2024-08-29
[build] [deps] Bump ppxlib toolchain to 0.26.0.

ejgallego:
It is time to clean up this old compat setup once that #698 has been completed; old `ppx_import` is giving us too much trouble with `petanque`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,11 @@
+# unreleased
+------------
+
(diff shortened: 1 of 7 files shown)


ejgallego:
@Alizter , how can we bump the Nix version of `ppx_import` ?

Versions older than 1.11 are giving some trouble, and they are pretty hard to test due to very old toolchain, so I'm updating the lower bounds (anyways OCaml < 4.14 is not supported anymore, so we should drop it at some point)

PR #815, opened 2024-08-30
[ci] [windows] Try OPAM 2.2 for the Windows build

ejgallego:
PR works, however we will wait for new releases of dune / ocamfind before merging this.


ejgallego:
Incorporated in #890 as the we had to bump the setup-ocaml action anyways as v2 was broken.

PR #823, opened 2024-09-18
[goals] Allow `command` to process several Coq commands.

ejgallego:
This is a prototype extension.

Note the important change here in behavior here, but IMHO it is
needed: if the commands in `command` fail, we will now fail the
request.

I think in general we may allow some form of error recovery here, but
for now we just let clients recover.

It does seem that CoqPilot / Pytanque do need even a more powerful
interface, but this will allow the them to make some progress for nwo.

Thanks to CoqPilot devs for feedback on this.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -40,6 +40,12 @@
    support Coq's 8.21 `quickFix` data (@ejgallego, #840, #843, #845)
  - [petanque] Fix bug for detection of proof finished in deep stacks
    (@ejgallego, @gbdrt, #847)
+ - [goals request] allow multiple Coq sentences in `command`. This is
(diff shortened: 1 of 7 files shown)


ejgallego:
cc: @amblafont , note the two changes in goal pre-processing:
- we now admit more than one command
- if the command fails, the request will fail

AFAICT they should be backwards compatible with your use case.

We will provide a more advanced speculative execution API soon, however I don't think you need it.

ejgallego:
@gbdrt this is the PR implementing `run`, note how the code duplication goes away, and now the API is properly set in the `Fleche.Doc` module.

This way we can access the more fancy document processing stuff, like error recovery, from run finally!

PR #829, opened 2024-09-25
[plugins] Example for error explaining plugin.

ejgallego:
Inspired from the comments on Coq's issue 19601.

For example, for this file:
```coq
Definition a := 3.

Lemma foo (n : nat) (f : nat -> nat) : True = False.
Proof.
now reflexivity.
set (g := 3).
Qed.
``

ẁe get:

```
[message] [explain errors plugin] we got 2 errors
[message] [explain errors plugin] Error:
 In environment
 n : nat
 f : nat -> nat
 Unable to unify "False" with "True".
for goals:

     n:nat
     f:nat -> nat
     ----------------
     True = False

[message] [explain errors plugin] Error:
  (in proof foo): Attempt to save an incomplete proof
 (there are remaining open goals).
for goals:

     n:nat
     f:nat -> nat
     g:nat
     ----------------
     True = False

```


ejgallego:
Ok, added the command that failed to the output, IMHO this is good for merge.

We could add some better bells and whistles, for example using a library to highlight the command better on the console, as most compilers do.

PR #834, opened 2024-09-27
Utils.with_control use VernacControl (adapt to coq/coq#19517 adding a control flag)

SkySkimmer:


coq/utils.ml
```diff
@@ -45,47 +45,14 @@ let to_range ~lines (p : Loc.t) : Lang.Range.t =
...
-  | Vernacexpr.ControlFail ->
-    with_fail fn;
-    Vernacstate.Interp.invalidate_cache ();
-    Vernacstate.unfreeze_full_state st
-  | Vernacexpr.ControlSucceed ->
-    fn ();
-    Vernacstate.Interp.invalidate_cache ();
-    Vernacstate.unfreeze_full_state st
-  (* Unsupported by coq-lsp, maybe deprecate upstream *)
-  | Vernacexpr.ControlRedirect _ -> fn ()
```
SkySkimmer, line 88: Why is redirect unsupported? Should I filter them out before sending the control flags to VernacControl?

ejgallego: I guess no real reason, mainly because it is a kind of security hole, and indeed, you'd like the document manager or build system to be notified that a command is going to write a file.

But IMHO things are fine as they are, if people use `Redirect` in `coq-lsp` they will get files written in their FS, their choice I guess.

SkySkimmer:
This is backwards compatible AFAICT so please merge.

PR #839, opened 2024-09-28
[hover] Fix universe printing in hover.

### LINKED ISSUE #835: Unvierse levels displayed as de bruijin indices in hover
When hovering over a term with `Set Printing Universes`, we expect to see nicely named universe variables.
```coq
From Coq Require Import Prelude.
Set Printing Universes.
Set Universe Polymorphism.

Inductive foo : Type := bar : Type -> foo.

Print foo.
```

Here `Print` outputs the following:
```coq
Inductive foo@{u} : Type@{u+1} :=  bar : Type@{u} -> foo@{u}.
(* u |=  *)

Arguments bar _%type_scope
```
Which is reasonable. However hovering over `foo` outputs:
```coq
foo : Type@{Var(0)+1}
```
which is not very readable. Since `About` and `Print` are able to provide reasonable terms, it seems there is some configuration missing in the term printer.

ejgallego:
Our code was incomplete. The API here could be really improved Coq-side. About is too verbose for hover, hence our code. We could opt for the version in `prettyp.ml` tho.

Fixes #835

### DIFF
--- a/controller/rq_hover.ml
+++ b/controller/rq_hover.ml
@@ -12,10 +12,28 @@ let build_ind_type mip = Inductive.type_of_inductive mip
 
 type id_info =
   | Notation of Pp.t
-  | Def of (Pp.t * Names.Constant.t option * string option)
+  | Def of
+      { typ : Pp.t  (** type of the ide *)
+      ; params : Pp.t  (** params that need display next to the name *)
+      ; full_path : Names.Constant.t option
+            (** full path of the constant, if any, for example
+                [Stdlib.Lists.map] *)
+      ; file : string option  (** filename where the constant is located *)
+      }
 
-let info_of_ind env sigma ((sp, i) : Names.Ind.t) =
+let print_params env sigma params =
+  if CList.is_empty params then Pp.mt ()
+  else Pp.(spc () ++ Printer.pr_rel_context env sigma params ++ brk (1, 2))
+
+let info_of_ind env ((sp, i) : Names.Ind.t) =
+  let udecl = None in
   let mib = Environ.lookup_mind sp env in
+  let bl =
+    Printer.universe_binders_with_opt_names
+      (Declareops.inductive_polymorphic_context mib)
+      udecl
+  in
+  let sigma = Evd.from_ctx (UState.of_names bl) in
   let u =
     UVars.make_abstract_instance (Declareops.inductive_polymorphic_context mib)
   in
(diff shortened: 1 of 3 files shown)


Alizter:
How does the new version print?

controller/rq_hover.ml
```diff
@@ -128,10 +162,11 @@ let pp_file fmt = function
   | Some file -> Format.fprintf fmt " - **in file**: `%s`" file
 
 let pp_typ id = function
-  | Def (typ, cr, file) ->
+  | Def (typ, param, cr, file) ->
```
Alizter, line 165: Probably best to call this `sort_param` or something.

ejgallego:
```coq
foo@{u u0 u1} (M : Type@{u}) : Type@{u0} -> Type@{max(u,u1+1)}
```

controller/rq_hover.ml
```diff
@@ -12,10 +12,21 @@ let build_ind_type mip = Inductive.type_of_inductive mip
 
 type id_info =
   | Notation of Pp.t
-  | Def of (Pp.t * Names.Constant.t option * string option)
+  | Def of (Pp.t * Pp.t * Names.Constant.t option * string option)
```
Alizter, line 15: Might be a good idea to make this a record and give it some proper field names. It would make it easier to read and understand.

ejgallego: All the code here was mostly placeholder as to test hover and wasn't properly designed, however it seems we got stuck with it for a while, due to other priorities.

If you look at the code, it is indeed pretty horrible haha. The Coq API doesn't absolutely help tho.

Alizter: Yeah, we should really try to upstream some better API changes on the Coq side. I believe it isn't considered a priority due to the mostly internal nature.

Alizter (approved):
Seems like a fine patch for user experience, but the API could use a cleanup here and upstream.

controller/rq_hover.ml
```diff
@@ -12,10 +12,28 @@ let build_ind_type mip = Inductive.type_of_inductive mip
...
-  | Def of (Pp.t * Names.Constant.t option * string option)
-
-let info_of_ind env sigma ((sp, i) : Names.Ind.t) =
+  | Def of
+      { typ : Pp.t  (** type of the ide *)
+      ; params : Pp.t  (** params that need display next to the name *)
+      ; full_path : Names.Constant.t option
+            (** full path of the constant, if any, for example
+                [Stdlib.Lists.map] *)
+      ; file : string option  (** filename where the constant is located *)
```
Alizter, line 21: Do we have access to the location too? (The one goto definition uses). Not relevant for this PR, but it might be useful to include.

ejgallego: We could, but not sure line number is relevant here, the client can indeed issue the corresponding specific request for location.

PR #843, opened 2024-09-30
[lang] Move `Lang.Diagnostics.Data.t` from a list to name field

ejgallego:
This fits our use case better.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -34,6 +34,8 @@
    info that is needed is at hand. It could also be tried to set the
    build target for immediate requests to the view hint, but we should
    see some motivation for that (@ejgallego, #841)
+ - [lsp] [diagnostics] (! breaking change) change type of diagnostics
+   extra data from list to named record (@ejgallego, #843)
 
 # coq-lsp 0.2.0: From Green to Blue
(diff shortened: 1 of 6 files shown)


CHANGES.md
```diff
@@ -34,6 +34,8 @@
    info that is needed is at hand. It could also be tried to set the
    build target for immediate requests to the view hint, but we should
    see some motivation for that (@ejgallego, #841)
+ - [lsp] [diagnostics] (! breaking change) change type of diagnostics
+   extra data from list to named record (@ejgallego, #843)
```
Alizter, line 38: I think these should be in a separate subsection of the changelog near the top with the title "Breaking changes".

ejgallego: Indeed, we haven't been too formal as we considered the 0.1.x series not stable at all, we can try to improve in the 0.2.x series, but I think 0.3.x (which could become a kind of 1.0 beta) could enforce this more strictly.

ejgallego: I noted this in the changelog because I know water proof is using `sentenceRange`. Also because this change was important moving forward with stuff like #840

PR #859, opened 2024-10-17
Adapt to coq/coq#19697 (genarg register gives ArgT.tag to "default")

SkySkimmer:


ppedrot:
Please merge now.

ejgallego:
> In my case, I just don't know what the process is to actually merge, as the "merge pull request" button is grayed out on the GitHub page.

Does it say why?

You should have merge rights, the process is indeed to push the merge button (tho we appreciate if devs update the coq submodule so the CI can also be checked)

SkySkimmer:
It looks like this
Screenshot_20241018_152948
(on https://github.com/ejgallego/coq-lsp/pull/810)

ejgallego:
Ok so it is the signing, thanks for the info. If you sign your commits you should be able to merge.

I will raise this topic on a call, I'd be happy to remove the signing requirement but indeed there are some legal ramifications that we may want to discuss; based on what I know maybe instead or removing the signing requirement here we would like to require a `Signed-off` field for Coq contributions.

SkySkimmer:
>tho we appreciate if devs update the coq submodule so the CI can also be checked

That doesn't sound very practical before the upstream PR is merged, if we update the submodule to its current commit it may get out of date when the upstream PR gets rebased or otherwise changed

ejgallego:
Indeed doing this like that would not be practical, IMHO we should figure out some kind of automatic way to do this, so overlay CIs can somehow detect the Coq version they should run CI against when building an overlay in their CI.

ppedrot:
I activated the default signing, let's see what it gives on the next overlay.

PR #863, opened 2024-10-18
Adapt to coq/coq#19726 (removed some unused genargs)

SkySkimmer:


ejgallego:
This seems backwards compatible so merging.

Issue #854, opened 2024-10-07
Syntax highlighting


Alizter:
Various vernac commands are highlighted incorrectly. It's difficult to be exhaustive, so we should keep this issue open to track them. Here are the ones I've come across so far:

- [ ] `Declare ML Module`
- [ ] `Set Default Proof Mode`
- [ ] `Set Strict Universe Declaration`
- [ ] `Global Set Default Goal Selector "!"` highlights `Goal` and `Selector` incorrectly separately
- [ ] `Create HintDb`
- [ ] `Register`
- [ ] `Add Search Blacklist`
- [ ] `Add Printing Let`
- [ ] `Hint Unfold`
- [ ] `Scheme`
- [ ] `Opaque`
- [ ] `Monomorphic` modifier
- [ ] `Polymophic` modifier

ejgallego:
This is something we inherit from C.J.'s VsCoq 1, I think that VSCoq 2 also uses the same file; maybe we could sync the effort on Zulip to improve this.

Of course we can add the LSP highlight method (semanticsToken) to coq-lsp, I am not sure about what the mapping for Coq would be.

4ever2:
> This is something we inherit from C.J.'s VsCoq 1, I think that VSCoq 2 also uses the same file; maybe we could sync the effort on Zulip to improve this.

VSCoq 2 also inherited the same file from VSCoq 1, however, all three versions have diverged since then. Interestingly VSCoq 1 seems to have the most up-to-date syntax grammar.

I have a local version of the grammar that fixes several of the issues mentioned and should be mostly up-to-date with Coq 8.20. I'd be happy to open a PR with the changes if you want.

ejgallego:
Hi @4ever2 , indeed it hasn't been easy to coordinate that; I'd be amazing if you would open a PR updating our definitions!

Please don't forget to update the changelog so your contribution is credited properly.

ejgallego:
@Alizter we can also implement semantic tokens, that should be easy for commands that are not defined in plugins.

ejgallego:
In order to implement semantic tokens in coq-lsp we need to:
- add the different kind of tokens and types to `lang`, maybe to `lang/ast.mli` ?
- write a function in `coq`, maybe in `coq/ast.mli` that maps Rocq vernaculars to the semantic tokens information.
- implement the semantic token request, add `controller/rq_stoken.ml`, plug it in `lsp_core.ml` and add the capability in `controller/rq_init.ml`.

[...]

PR #872, opened 2024-11-13
[vscode] Syntax highlighting for Coq 8.17-8.20

4ever2:
Add updated syntax from VSCoq Legacy and extend with new commands from Coq 8.17-8.20

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 
  - [fleche] fix quick fixes for errors being lost due to incorrect
    handling of `send_diags_extra_data` (@ejgallego, #850)
+ - [vscode] Syntax highlighting for Coq 8.17-8.20 (@4ever2, #872)
 
 # coq-lsp 0.2.2: To Virtual or not To Virtual
 ---------------------------------------------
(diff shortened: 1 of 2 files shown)


Alizter:
Did you cover the list in #854? I haven't been able to fully check myself yet.

4ever2:
Not all of them are covered by this PR. Some of them seem quite hard to fix.
I don't think there is a good way to fix highlighting on `Set`.

Alizter:
I guess to do `Set` correctly you will have to look at all the options. IIRC there should be a vernac command that prints the options table, so that might be a way to start on such a list.

I can't tell if `Register some_constant.` will get highlighted, I can see `Register Scheme` is picked up.

Also, I can't tell if `Add Search Blacklist` is also considered.

I've updated the list accordingly since presumably we will merge this anyway.

4ever2:
`Register some_constant.` will highlight `Register` as a command but not `some_constant` as an ident.
`Add Search Blacklist` is not considered here.

Alizter:
> `Register some_constant.` will highlight `Register` as a command but not `some_constant` as an ident.

That's fine, as long as `Register` is being picked up.

PR #873, opened 2024-11-15
Adapt to coq/coq#19834 (removal of Loadpath.has_ml field)

SkySkimmer:
Although this removes instances of `has_ml = true`, `has_ml` is only used to send the path to the `add_dir` of the registered toplevel in Mltop. For coq-lsp the `has_dir` is noop (`coq/init.ml` line 76). So there should be no change in behaviour.


ejgallego:
Thanks!

PR #874, opened 2024-11-15
[plugins] Fix the output of the AST dump plugin to be valid JSON

blackbird1128:
This PR fix the output produced by the AST dump plugin to output valid JSON.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 # unreleased
 ------------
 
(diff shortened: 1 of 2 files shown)


blackbird1128:
Hello, no problem !
Yes that a good idea, I can make the change if you want, tell me.

Yes, this is also fine, I can also sign commits if needed will try to do so in the future, not sure about how to do that retro-actively tho.

PR #884, opened 2024-12-17
Adapt w.r.t. coq/coq#19943.

ppedrot:


ppedrot:
This is not super pretty given that sexplib doesn't like GADTs. I had to hack my way around.

palmskog:
@ejgallego personally I think it would be nice to investigate kernel-level fuzzing (the original idea), but there are no plans from my side (or any other side I know about) to pursue this right now.

ejgallego:
> @ejgallego personally I think it would be nice to investigate kernel-level fuzzing (the original idea), but there are no plans from my side (or any other side I know about) to pursue this right now.

The original infra is still functional, and the fuzzers are much better these days, it wouldn't be too hard to co-advise an internship on that topic (I can get students / funds in Paris).

palmskog:
> The original infra is still functional, and the fuzzers are much better these days, it wouldn't be too hard to co-advise an internship on that topic (I can get students / funds in Paris).

Then sure, if you find a student (and you have funding for your own time), please let me know.

PR #887, opened 2025-01-09
Adapt w.r.t. coq/coq#19995.

ppedrot:
I kept the module representations opaque, I doubt that people rely on this through serlib given the mess that this data structure is.


ejgallego:
Thanks @ppedrot , the original motivation to expose this was to build a kernel fuzzer, actually that's still an interesting project, let me know if you are interested.

PR #889, opened 2025-01-15
Remove a now internal type from Declarations.

ppedrot:
This type is only used internally to the module implementation and is bound to be removed from the API in coq/coq#20061.


ppedrot:
This is backwards compatible, is this fine if I merge it @ejgallego ?

ejgallego:
Let me fix the CI, just in case.

ejgallego:
@ppedrot , PR is fine, will take care of it myself once I sort out a couple of things more on the infra.

PR #891, opened 2025-01-16
Adapt to coq/coq#19901 (relocatable mode)

SkySkimmer:
This doesn't feel very satisfying but IDK what else to do.

### DIFF
--- a/petanque/json_shell/shell.ml
+++ b/petanque/json_shell/shell.ml
@@ -4,8 +4,23 @@ let init_coq ~debug =
   let vm, warnings = (true, None) in
   Coq.Init.(coq_init { debug; load_module; load_plugin; vm; warnings })
 
+(* can't use Boot.Util.relocate because we're not together with rocq.exe *)
+let conf_coqlib =
+  match Coq_config.coqlib with
+  | NotRelocatable p -> p
+  | Relocatable p ->
+    let paths = match Sys.getenv_opt "PATH" with
+      | None -> []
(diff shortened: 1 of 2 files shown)


petanque/json_shell/shell.ml
```diff
@@ -4,9 +4,24 @@ let init_coq ~debug =
...
+        String.split_on_char sep paths
+    in
+    let name = if Sys.(os_type = "Win32" || os_type = "Cygwin") then "rocq.exe" else "rocq" in
+    let bindir = fst (System.find_file_in_path ~warn:false paths name) in
+    Filename.concat (Filename.concat bindir "..") p
+
 let cmdline : Coq.Workspace.CmdLine.t =
-  { coqlib = Coq_config.coqlib
-  ; coqcorelib = Filename.concat Coq_config.coqlib "../rocq-runtime"
+  { coqlib = conf_coqlib
```
ejgallego, line 23: Seems to me like this should be the same time than the one in coq_config, but see my comment in the original PR.

ejgallego: Sorry it wasn't sent yet went I wrote this. I'm happy consuming the new variant type, and following the spec at my side. I'm also happy for coq-lsp to error on relocatable so we don't block on coq-lsp implementing the spec, but we need a spec.

ppedrot:
@SkySkimmer you need to sign the commits for this to be mergeable.

PR #895, opened 2025-01-28
Stop relying on environ internals.

ppedrot:
More than half of the fields exposed in the private Environ.env module are not supposed to be accessed by anybody outside of the implementation. Some are actually just serialized to opaque data by Serlib.

Instead of exposing all this crap to the serialization process, we go through a better behaved type that only exposes the bits that are supposed to be accessed.

### DIFF
--- a/serlib/ser_environ.ml
+++ b/serlib/ser_environ.ml
@@ -73,9 +73,28 @@ module Globals = struct
   include SerType.Pierce(PierceSpec)
 end
 
-type env =
-  [%import: Environ.env]
-  [@@deriving sexp_of]
+module InternalEnv =
+struct
+  type _t = {
+    env_globals       : Globals.t;
+    env_named_context : Constr.named_context;
+    env_rel_context   : Constr.rel_context;
+    env_universes : UGraph.t;
+    env_qualities : Sorts.QVar.Set.t;
+    env_typing_flags  : Declarations.typing_flags;
+  } [@@deriving sexp_of]
+
+  let of_t env = {
+    env_globals = env.Environ.env_globals; (* TODO: use an accessor *)
+    env_named_context = Environ.named_context env;
+    env_rel_context = Environ.rel_context env;
+    env_universes = Environ.universes env;
+    env_qualities = env.Environ.env_qualities; (* TODO: use an accessor *)
+    env_typing_flags = Environ.typing_flags env;
+  }
+end
+
+type env = Environ.env
 
 let env_of_sexp = Serlib_base.opaque_of_sexp ~typ:"Environ.env"
 
@@ -83,7 +102,7 @@ let abstract_env = ref false
 let sexp_of_env env =
   if !abstract_env
   then Serlib_base.sexp_of_opaque ~typ:"Environ.env" env
-  else sexp_of_env env
+  else InternalEnv.sexp_of__t (InternalEnv.of_t env)
 
 type ('constr, 'term) punsafe_judgment =
   [%import: ('constr, 'term) Environ.punsafe_judgment]


ppedrot:
Note that this should even be more efficient in practice. A good chunk of the data from environ is duplicate redundant, being only there for algorithmic purposes (e.g. representing contexts both as lists and binary trees) so that when serializing one pays twice the price, but the algorithmic part is useless since serializing is already at least linear in the size of the structure.

ejgallego:
This is a double-edged sword, serialization is not about efficiency but about being faithful to the underlying structure. The motivation seems vague to me, sorry. A few questions:

- Why is this refactoring not done upstream?
- What does "private" in  "private `Environ.env` module" mean?
- What is the definition of "better behaved"?

I am not against this per-se, however introducing this new types adds to the maintenance burden of serlib so I'd like to understand the benefits first. (For example now I have to care whether a new relevant field to env has been added)

serlib/ser_environ.ml
```diff
@@ -73,17 +73,36 @@ module Globals = struct
...
-  [@@deriving sexp_of]
+module InternalEnv =
+struct
+  type _t = {
+    env_globals       : Globals.t;
+    env_named_context : Constr.named_context;
+    env_rel_context   : Constr.rel_context;
+    env_universes : UGraph.t;
+    env_qualities : Sorts.QVar.Set.t;
+    env_typing_flags  : Declarations.typing_flags;
```
ejgallego, line 84: Adding to my point in the comment, it seems this structure is already missing two fields that are relevant, in particular `rewrite_rules_allowed` and `symb_pats`, likely `retroknoledge` too?

ppedrot: Retroknowledge is garbage for users, this is kernel-facing and (unfortunately) mostly unsound.

As for rewrite rules, the implementation is going to change anyway, and good luck figuring out what to do with the data there. I'm not sure what the point of exposing all of this is. Blindly serializing internal data without even checking that the underlying types have changed is looking for trouble. That's not an API, that's barely better than a memory dump and let the end-user figure out what the pointer addresses are.

ppedrot:
On a more abstract level, and forgetting locally about the "ser" in serlib,  I don't quite understand this insistence on serializing the environment type. An environment is like a database, and exposing an API to a database that requires the user to dump the entire data in a textual format is never going to scale. There should be instead a notion of cursor or something like that to query the data there. We could call it, I don't know, a kernel name?

ppedrot:
Ping @ejgallego. The alternative is to expose a view type for Environ.env, just like we do for Globals. This makes the code on the serlib side more robust, but it shifts the control of the interface to the Rocq side. I can try to submit this as an alternative PR + overlay.

ppedrot:
See coq/coq#20165 and #897 for an alternative.

ppedrot:
I'm closing this one in favour of https://github.com/ejgallego/coq-lsp/pull/897.

ejgallego:
Thanks for the patch @ppedrot ,

> Last case from yesterday: messing with delta-resolvers. Ironically enough, this precise endeavour was already mentioned in [coq/coq#7745](https://github.com/coq/coq/pull/7745). Given the time passed since then and the lack of progress on the topic, I take this as a constructive evidence that exposing the internals in the environment type is _really_ slowing down progress.

I am sorry, but I really fail to see how this minor detail upstream has slowed down progress on delta-resolvers, I don't follow the reasoning here (and indeed after re-reading 7745 it actually gotta stand by the position we took there)

In particular, I think we should separate policy from technical style points; these are separate matters that need to be evaluated separatedly, even if they interact non-trivially often.

> On a more abstract level, and forgetting locally about the "ser" in serlib, I don't quite understand this insistence on serializing the environment type. An environment is like a database, and exposing an API to a database that requires the user to dump the entire data in a textual format is never going to scale. There should be instead a notion of cursor or something like that to query the data there. We could call it, I don't know, a kernel name?

I agree that the original use case for this patch (by Kaigu) was pretty weak, I did recommend them going a different route, which they do now (including having migrated to Lean)

On the other hand, while I understand your point of view, it is beyond the scope of what serialization does. Serialization is not an API, actually it is the dual of it. In particular:
- serlib doesn't provide any kind of stability guarantee, on the contrary, it actively states that the format is internal (which doesn't matter a lot these days, but that's a different question)
- serlib doesn't try to be smart, but faithful: this is great for some use cases (think of debugging), bad for others. Seeing something closer to what Coq really represents has been a very interesting experience often.

The main problem why kernel env seralization doesn't work is different tho: sharing is lost, and the traces are unusable.

We could remove serialization of kernel envs if we'd like to IMHO. It's been there mostly as a sanity check as to ensure we can still serialize enough to build a kernel trace, but that's all.

PR #926, opened 2025-03-07
VSCode extension keybindings shouldn't conflict with editor defaults.

listochkin:
On macOS `Cmd+N` is used for creating a new file, `Cmd+P` is "Go to file ..." - both are fundamental editor features, - that this extension breaks by introducing conflicting key binds.

macOS doesn't have a `Meta+` key prefix [(You can see the available prefixes here)](https://code.visualstudio.com/docs/editor/keybindings#_accepted-keys). `Alt+Enter` is used for "select all" during search and for other cases here and there. In addition, it's used by a popular GitLens extension. I propose `Alt+Shift+enter` as a new key combination, at least on macOS.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -50,6 +50,9 @@
  - [doc] [faq] Updated FAQ to account for VSCoq 2 release in 2023,
    thanks to Patrick Nicodemus for pointing out the outdated
    documentation (@ejgallego, #846, fixes #817)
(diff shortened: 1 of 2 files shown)


ejgallego:
Thanks a lot @listochkin , and sorry for the delay, can you add an entry in the CHANGEs.md file? Happy to do it myself.

ejgallego (approved):

PR #946, opened 2025-04-16
Changed bound_names in rocq

LeoAlexElouan:


serlib/ser_uvars.ml
```diff
@@ -79,11 +82,10 @@ end
 
 module AbstractContext = struct
 
-  let hash_fold_array = hash_fold_array_frozen
```
SkySkimmer, line 82: Is this change actually necessary? What's going on here?

LeoAlexElouan: When the change below this one is done, this function becomes unused.
I think maybe [@@deriving hash] worked with the array product, but not the record of two arrays.

SkySkimmer (approved):

PR #949, opened 2025-05-06
Adapt to rocq-prover/rocq#20605 (cleaned up projection flags)

SkySkimmer:


SkySkimmer:
Is CI not using rocq dev? coq_lsp is passing in master https://gitlab.inria.fr/coq/coq/-/jobs/5706390

SkySkimmer:
This change may be backwards compatible (the removed type hasn't been used in the parser produced ast for a while AFAIK), let's see if rerunning CI after https://github.com/ejgallego/coq-lsp/pull/950 works

PR #960, opened 2025-05-19
[petanque] Add feedback field to errors and `Run_result.t`

ejgallego:
This is important for many use cases. We also return feedback on petanque errors.

cc: @JulesViennotFranca


JulesViennotFranca:
That's really nice ! Thanks a lot @ejgallego :smile:

PR #993, opened 2025-07-12
Configurable completion chars

Durbatuluk1701:
This PR adds support for configuring the set of commit characters used in unicode completion within the Coq LSP server and VSCode extension.

### Motivation

Previously, the set of characters that would accept/commit a unicode completion was hardcoded.
The hardcoding of `" "` as a commit char was especially problematic for me, as typing conjunction (`/\`) followed by a spaced would always end up changing it to some unicode completion.

### Changes

- Added a new server configuration option to specify the commit characters for unicode completion.
- Updated the LSP and VSCode extension to respect this configuration.

### Impact

- Customizable unicode completion.
- (Hopefully) No breaking changes; default behavior remains unchanged unless the option is set.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,6 @@
 unreleased
 ----------
 
- - [lsp] [vscode] Unicode completion commit characters are now configurable via server settings. (@Durbatuluk1701)
  - [js] [deps] Bump to findlib 1.9.8, use vanilla API for loading and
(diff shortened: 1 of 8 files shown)


ejgallego (approved):
LGTM thanks!

Indeed, LSP doesn't provide a super way to control this stuff.

Note: I will move the completion configuration to its own section before the next release, so the setting will be under a `.completion` prefix.

Issue #941, opened 2025-04-10
`error` field of `proof/goals` does not respect `goal_after_tactic`


jpoiret:
**Describe the bug**
If `goal_after_tactic` is set to `false`, the `error` field of the `proof/goals` request will contain the error message of the sentence under the requested position (seemingly executing it?) instead of the last sentence.

**Desktop (please complete the following information):**
- `coq-lsp 0.2.2-8.20.0`
- using my own `rocq-mode.el` on Emacs

ejgallego:
Thanks @jpoiret , indeed the logic for goals and errors is different. I think here you may be witnessing 2 things:
- over-execution: in 0.2.2 we detected some commands do over-execute, this will be fixed in 0.2.3
- the logic for errors don't respect the attribute

So indeed, these could be two issues that interact but are different.

Can you provide an example .v Rocq file so we can reproduce the problem and add a test to the test-suite? Thanks!

ejgallego:
Ping @jpoiret , I want to check that my understanding of this issue is correct:

I guess you want a fully lazy mode, so that sentences are not executed while you are writing it, correct?

In the first LSP prototypes, the idea is that the client would only issue the goals request when it wanted to display the goals, for example if you are in "on-demand mode", when the user press "Ctrl+Enter" (display goals), but I can see how this is inconvenient.

The behavior here is subtle so I'd like to be sure the fix captures your workflow properly.

jpoiret:
I don't think the problem really is about execution, but rather the contents of the returned response: the `error` field of a `proof/goals` request should return the errors of the sentence that the `proof/goals` request targets, for consistency, rather than always the one under point.

Wrt. a fully lazy mode, I believe lsp-coq already supports it with the check on request mode. It just is surprising that when requesting a goal display for the sentence before the one under point, the sentence under point would also get executed.

jpoiret:
As for a sample `.v` file, the following is sufficient to demonstrate the behavior:
```
Goal True -> False.
  intros ?.
  constructor.
```

If you have `goal_after_tactic` set to `false`, and request `proof/goals` with the point on the first `t` of `constructor`, you will get `The type has no constructors.` in the `error` field.

PR #999, opened 2025-07-24
[lsp] Return executed sentence info in goal request.

jpoiret:
All info returned by a `proof/goals` request should be about the sentence requested by the user, whereas currently only `goals` follows this, and `messages`, `error` and `range` are about the sentence under point instead. IMO this is bad UX and may mislead users, as the goal display is de-correlated from the associated info and error messages and range indicators.

There is no loss of information here, as anyways diagnostics are sent back with their range info using the usual LSP mechanisms, and the user can peruse those under point using the standard method of their editor.

WDYT?

PS: I didn't put anything under the changelogs because I don't know what the policy is esp. regarding versioning (this is a breaking change).

fix: #941

### DIFF
--- a/etc/doc/PROTOCOL.md
+++ b/etc/doc/PROTOCOL.md
@@ -610,6 +610,15 @@ the `initializationOptions` parameter for the LSP `init` method.
 As of today, the server exposes the following parameters:
 
 ```typescript
+export interface UnicodeCompletionConfig {
+    enabled: "off" | "normal" | "extended";
+    commit_chars : string[];
+}
+
+export interface CompletionConfig {
+    unicode: UnicodeCompletionConfig
+}
+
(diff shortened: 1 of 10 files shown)


etc/doc/PROTOCOL.md
```diff
@@ -280,7 +280,11 @@ interface GoalRequest {
 The `textDocument` and `position` parameters are standard.
 
 Note that `rocq-lsp` will execute the Rocq document up to the
-`position` specified in the request paremeters.
+sentence specified in the request by the `position` and `mode` paremeters.
```
ejgallego, line 283: Note: `mode` doesn't control execution.

ejgallego:
Ouch, merged in mistake, but see latest commit for changes I did on top of the PR, let me know how they do look.

PR #1004, opened 2025-08-05
Remove Base

patrick-nicodemus:
Looks like the Jane Street Base library accidentally made its way into the Fleche.Doc module. This function is now in the stdlib, so it's unnecessary to rely on Base here.


patrick-nicodemus:
Oh, I get it now.
You are using `Base` as a dependency, but that is not explicit in the `opam` description, it is required transitively because you rely on ppx's which require `Base`. I am using a slightly more recent version of the ppxes where the dependency on `Base` has been decoupled, so I don't have `Base` installed and my compilation fails.
So, either `Base` should be added explicitly to the opam dependencies or `Base` should be removed from the codebase.

controller/nt_cache_trim.ml
```diff
@@ -48,7 +48,12 @@ let caches () =
 let pp_cache fmt (name, freqs) =
   let zsum = List.filter (Int.equal 0) freqs in
   let pp_zsum fmt l = Format.fprintf fmt "{ 0-entries: %d }" (List.length l) in
-  let fsum = Base.List.take freqs 20 in
+  let fsum = (* List.take added in OCaml stdlib 5.3 *)
+    (let rec take n l = if n = 0 then [] else
+        match l with hd :: tl -> hd :: take (n-1) tl | [] -> []
+    in
+    take 20 freqs
+   ) in
```
ejgallego, line 56: You can use `CList.firstn` here if you want.

coq-lsp.opam
```diff
@@ -19,7 +19,7 @@ doc: "https://ejgallego.github.io/coq-lsp/"
 
 depends: [
 
-  ("ocaml" {>= "5.0"} | ("ocaml" {<= "5.0"} & "memprof-limits" { >= "0.2.1" } ))
+  ("ocaml" {>= "5.0"} | ("ocaml" {>= "4.05"} & "memprof-limits" { >= "0.2.1" } ))
```
ejgallego, line 22: Why this change?

patrick-nicodemus:
> I was assuming that `Base` would be hard to avoid, for example, I see that `ppx_hash` and `ppx_compare` still depend on `base`, what versions of these are you using? I'd be important to ensure the CI is checking the build.

I am using a preview of the Jane Street suite v18, their preview is here. https://github.com/janestreet/opam-repository
I was curious about their OxCaml fork so I added their development opam repository to my local setup.

coq-lsp.opam
patrick-nicodemus: Okay. I removed this commit.

ejgallego (approved):

PR #1017, opened 2025-09-10
V8.20 more commands

JulesViennotFranca:
Hello @ejgallego, here is the pull request @gbdrt told you about. I added a way to list notations inside of a statement. It should be easy to extend this work to notations inside of `Definition` and `Goal` as well. If it could even be improved to definition of notations on hover for coq-lsp that would be incredible !

### DIFF
--- a/petanque/agent.ml
+++ b/petanque/agent.ml
@@ -338,6 +338,20 @@ let simple_run_result res feedback =
   let feedback = List.map fb_print_string feedback in
   Run_result.{ st = res; hash; proof_finished; feedback }
 
+let list_notations_in_statement ~token ~st ~statement () :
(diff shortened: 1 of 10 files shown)


ejgallego:
Hi @JulesViennotFranca , thanks so much for the PR! A few notes:

- as in Rocq upstream, PRs need to target the `main` branch.
- as this seems to be using Rocq code, we'd need agreement from Rocq's team about a possible duplication and the impact on Rocq's CI
- I have a few comments to do about the code, I'll be happy to review once the PR is rebased for `main`

ejgallego:
Taking over, rebasing into main, and doing a bit of clean up.

PR #1053, opened 2025-10-31
Adapt to rocq-prover/rocq#21254 (stop catching async exns)

SkySkimmer:
Not sure how we should consider async exns in this code, for now I made them considered as anomalies.


SkySkimmer:
IDK if it works, but is_anomaly returned true on the memprof exceptions AFAIK so if it doesn't work then it also didn't work before

